PR #1, opened 2016-03-16
add state getter

jbwyme:


jbwyme:
i see this.state gets updated so no need!

PR #3, opened 2016-04-02
Provide onRendered view lifecycle hook

evnp:
I need a way to hook into the panel app update lifecycle, to execute code that manipulates the DOM after a view is rendered (specifically, to initialize MPPlatform widgets). This lets you define an `onRendered` method on any view that will be executed every time the view is rendered (after rendering is complete). Lmk if there's a better way to do this, this is just an idea.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -15,11 +15,11 @@ export default class App {
     this.state = initialState;
     this.router = new Router(this, options);
 
-    const screens = this.SCREENS;
-    for (let k in screens) {
-      screens[k].setApp(this);
+    this._screens = this.SCREENS;
+    for (let k in this._screens) {
+      this._screens[k].setApp(this);
     }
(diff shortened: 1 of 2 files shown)


tdumitrescu:
It's just not the way to do it. Remember that all `render` does is produce VDOM nodes which are then diffed against what's actually there for potential updates - `render` can be called any number of times without the DOM actually changing, and needs to remain idempotent. What you want is either web components (my pref) or `virtual-dom` widgets (less portable).

tdumitrescu:
(BTW for a concrete example, looks at how `query-console` instantiates and manipulates ACE editor, which, like current MPPlatform widgets, offers only an imperative interface)

evnp:
Right, web components, that makes way more sense. I was wanting something in the same vein as React's componentDidUpdate and didn't even think of that. Thanks for the info!

PR #5, opened 2016-04-09
Allow deeply nested objects to be returned from VIEWS

evnp:
I'd like to be able to do things like
```
  get VIEWS() {
    return {
      panes: [
        {header: new PaneAHeaderView(this), content: new PaneAContentView(this)},
        {header: new PaneBHeaderView(this), content: new PaneBContentView(this)},
      ],
    };
  }
```
so I can have templates like
```
ul.panes
  for pane in views.panes
    li.pane
      .header= pane.header(state)
      .content= pane.content(state)
```

### DIFF
--- a/lib/view.js
+++ b/lib/view.js
@@ -1,15 +1,14 @@
 function deepMapViews(views, func) {
-  if (typeof views.render === 'function') {
-    return func(views);
-  } else if (Array.isArray(views)) {
+  if (Array.isArray(views)) {
     return views.map(view => deepMapViews(view, func));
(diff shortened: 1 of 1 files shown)


tdumitrescu:
Thx, merged with some tests (https://github.com/mixpanel/panel/blob/ff1a6b637f45ed152777ffbbdf46f1fa98c7d633/test/view.js#L79-L179). If those test cases look like they're missing any critical behavior from this patch, please update.

evnp:
Nice, sorry I neglected to add tests myself. Those look great to me.

PR #6, opened 2016-05-26
Make it possible to call update on a component before it's attached

evnp:
This is useful when you want to make an update call from
attributeChangedCallback, since that can occur before the
attachedCallback is executed. Init state in createCallback and
use that to store updates made before attachedCallback.

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -17,7 +17,7 @@ export default class Component extends WebComponent {
       routes: {},
       template: () => { throw 'No template provided by Component subclass'; },
     }, this.config);
-    this.state = this.state || {};
+    this.state = {};
   }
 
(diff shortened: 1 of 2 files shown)


lib/component.js
```diff
@@ -85,7 +86,9 @@ export default class Component extends WebComponent {
   }
 
   update(stateUpdate={}) {
-    if (this.isPanelRoot) {
+    if (typeof this.app === 'undefined') {
```
tdumitrescu: let's just use an explicit prop that gets set during the `attachedCallback`, e.g., `if (this.initialized) {`

tdumitrescu:
Awesome. Thx for the thorough test cases!

PR #7, opened 2016-06-01
shouldUpdate() minimal implementation

tdumitrescu:
@evnp @chzhcn

### DIFF
--- a/test/test-component.js
+++ b/test/test-component.js
@@ -77,7 +77,7 @@ describe('Nested Component instance', function() {
   var el, childEl;
 
   context('before attached to DOM', function() {
-    beforeEach(function(done) {
+    beforeEach(function() {
       document.body.innerHTML = '';
       el = document.createElement('nested-app');
     });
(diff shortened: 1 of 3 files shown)


test/test-component.js
```diff
@@ -77,31 +77,20 @@ describe('Nested Component instance', function() {
   var el, childEl;
 
   context('before attached to DOM', function() {
-    beforeEach(function(done) {
+    beforeEach(function() {
       document.body.innerHTML = '';
       el = document.createElement('nested-app');
```
evnp: Think we actually need this from below as well
```
       window.requestAnimationFrame(function() {
         childEl = el.getElementsByTagName('nested-child')[0];
         done();
       });
```
Otherwise childEl may be undefined, as seems to be happening in the CI test runs

tdumitrescu: well actually in order to test any of this 'before attached to DOM' context at all you need to be able to run assertions at the point when the parent template has been rendered but the child's `attachedCallback` has not yet run, which I don't think is possible with these integration-style tests. We'll have to rewrite these with the test code manipulating the components directly (e.g., manually running the created and attached callbacks)

tdumitrescu:
alright @evnp, tests are doing ok, added a basic one for `shouldUpdate()` too

evnp:
Ahh, that makes sense. Interesting that this is a case where things can't be done immutably. I should've followed the precedent set below in update anyway, sorry about that. lgtm

PR #8, opened 2016-06-06
Enable use of Shadow DOM

ilyakamens:


lib/component.js
```diff
@@ -55,7 +56,8 @@ export default class Component extends WebComponent {
     Object.assign(this.state, newState);
 
     this.loop = MainLoop(this.state, this._render.bind(this), {create, diff, patch});
-    this.appendChild(this.loop.target);
+    this.el = this._config.useShadowDom ? this.createShadowRoot() : this;
```
tdumitrescu: `this.getConfig('useShadowDom')`

lib/component.js
```diff
@@ -68,6 +70,12 @@ export default class Component extends WebComponent {
     this.$panelParent && this.$panelParent.$panelChildren.delete(this);
   }
 
+  attributeChangedCallback() {
+    if (this === this.$panelRoot) {
```
tdumitrescu: `if (this.isPanelRoot) {`

tdumitrescu: probably also want to check `this.initialized`

test/test-component.js
```diff
@@ -86,6 +86,27 @@ describe('Simple Component instance', function() {
...
+    beforeEach(function(done) {
+      el.setConfig('useShadowDom', true);
+      document.body.appendChild(el);
+      window.requestAnimationFrame(function() {
+        done();
+      });
+    });
+
+    it('creates and uses a shadow root', function() {
+      expect(el.el).not.to.equal(el);
```
tdumitrescu: this would be a good spot for: `expect(el.shadowRoot).to.be.ok();`

test/test-component.js
```diff
@@ -86,6 +86,27 @@ describe('Simple Component instance', function() {
...
+    });
+
+    it('renders its template', function() {
+      expect(document.getElementsByClassName('foo')).to.have.lengthOf(0);
+      expect(el.children).to.have.lengthOf(0);
+      expect(el.shadowRoot).not.to.equal(null);
```
tdumitrescu: does this work for `undefined` etc?

test/test-component.js
```diff
@@ -86,6 +86,27 @@ describe('Simple Component instance', function() {
...
+      });
+    });
+
+    it('creates and uses a shadow root', function() {
+      expect(el.el).not.to.equal(el);
+    });
+
+    it('renders its template', function() {
+      expect(document.getElementsByClassName('foo')).to.have.lengthOf(0);
+      expect(el.children).to.have.lengthOf(0);
```
tdumitrescu: `to.have.length(0)`

test/test-component.js
```diff
@@ -86,6 +86,27 @@ describe('Simple Component instance', function() {
       });
     });
   });
+
+  context('when using shadow root', function() {
```
tdumitrescu: 'when using shadow dom'

test/test-component.js
ilyakamens: yes, but i think using `ok` would be better

test/test-component.js
ilyakamens: talked offline. `lengthOf` is the way to go

lib/component.js
```diff
@@ -68,6 +71,31 @@ export default class Component extends WebComponent {
     this.$panelParent && this.$panelParent.$panelChildren.delete(this);
   }
 
+  attributeChangedCallback() {
+    if (this.initialized && this.isPanelRoot) {
+      this._applyStyles();
```
tdumitrescu: probably don't want to call this unless the correct attribute has changed

test/fixtures/shadow-dom-app.js
```diff
@@ -0,0 +1,16 @@
+import { Component } from '../../lib';
+import h from 'virtual-dom/virtual-hyperscript';
+
+var ShadowDomApp = document.registerElement('shadow-dom-app', class extends Component {
+  get config() {
+    return {
+      css: 'color: blue;',
```
tdumitrescu: it seems like the `css` prop is useless unless `useShadowDom` is set - should we have some sort of runtime check that you're not trying to use it when `useShadowDom` is false?

test/test-component.js
```diff
@@ -86,6 +86,41 @@ describe('Simple Component instance', function() {
...
+      expect(document.getElementsByClassName('foo')).to.have.lengthOf(0);
+      expect(el.children).to.have.lengthOf(0);
+      expect(el.shadowRoot.children[1].className).to.equal('foo');
+    });
+
+    it('applies the styles', function() {
+      expect(el.shadowRoot.children[0].innerHTML).to.equal('color: blue;');
+    });
+
+    context('when applying override styles', function() {
```
tdumitrescu: can we have one test case where the override is already set before `attachedCallback` fires? e.g. `<my-widget style-override="background: red">`

ilyakamens: added

tdumitrescu:
merged https://github.com/mixpanel/panel/commit/c9abb0812ffed5ed047c788533edc68379f29880, merci!

PR #9, opened 2016-06-07
don't console.error if there are no compiled routes

ilyakamens:


lib/router.js
```diff
@@ -46,7 +46,7 @@ export default class Router {
 
   navigate(fragment, stateUpdate={}) {
     fragment = stripHash(fragment);
-    if (fragment === this.app.state.$fragment && !Object.keys(stateUpdate).length) {
+    if (!this.compiledRoutes.length || fragment === this.app.state.$fragment && !Object.keys(stateUpdate).length) {
```
tdumitrescu: I would rather just not initialize/compile the routes/router in the first place rather than running a non-functional router if no routes are defined.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "panel",
-  "version": "0.7.2",
+  "version": "0.7.3",
```
tdumitrescu: no need to do this in PRs

PR #10, opened 2016-06-10
pass through shadow root when looking for parent

jbwyme:


lib/component.js
```diff
@@ -41,6 +41,9 @@ export default class Component extends WebComponent {
       this.isPanelChild = true;
       // find $panelParent
       for (let el = this.parentElement; el && !this.$panelParent; el = el.parentElement) {
+        if (el.nodeType === 11) {  // 11 = DOCUMENT_FRAGMENT_NODE
```
tdumitrescu: what do you think about pulling this out into an actual `const DOCUMENT_FRAGMENT_NODE = 11;` at the top of the module?

tdumitrescu:
good catch. can we get a test?

test/test-component.js
```diff
@@ -110,6 +110,15 @@ describe('Simple Component instance', function() {
       expect(el.shadowRoot).to.be.ok;
     });
 
+    it('successfully finds the panel root', function() {
+      el.attachedCallback();
```
tdumitrescu: `el.attachedCallback` should already have been run by now because of the `beforeEach`. maybe we should enforce non-reentrance

test/test-component.js
```diff
@@ -110,6 +110,15 @@ describe('Simple Component instance', function() {
...
+      childEl = document.createElement('nested-child');
+      childEl.setAttribute('panel-parent', el.panelID);
+      el.el.appendChild(childEl);
+      childEl.attachedCallback();
+      expect(childEl.$panelRoot).to.equal(el);
```
tdumitrescu: So this tests a child element finding a parent with shadow dom. I'd say it's also worth testing that a child element with shadow dom successfully finds a parent, what do you think?

jbwyme: well it won't matter since the top level element of the `child` el that has shadow dom is still in the light dom, but we could add a test anyways to prevent some future changes from breaking that

tdumitrescu: Yeah I guess I'm thinking of the high-level integration-test view where you have `el1 -> el2 -> el3` and if `el2` uses shadow dom, you want `el3` to keep finding `el1` as panel root no matter how our implementation changes.

PR #11, opened 2016-07-01
Safer rendering

tdumitrescu:
Changes some behavior around initialization and rendering so that template-rendering exceptions don't break the component. Fixes existing bad behavior:

- a rendering exception during the creation of the loop object would
throw synchronously and leave the component in a half-initialized
state (`this.loop` undefined, `this.initialized` not true, but element
attached to DOM nonetheless).

- a rendering exception during subsequent updates could lead to
main-loop's dreaded "Unexpected update occurred in loop" in
future passes because it didn't clean up its internal mutex.

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -72,13 +72,12 @@ export default class Component extends WebComponent {
 
     this.loop = MainLoop(this.state, this._render.bind(this), {create, diff, patch});
     this.el.appendChild(this.loop.target);
+    this.initialized = true;
 
     if (Object.keys(this.getConfig('routes')).length) {
       this.router = new Router(this, {historyMethod: this.historyMethod});
       this.navigate(window.location.hash);
     }
-
-    this.initialized = true;
   }
 
   detachedCallback() {
--- a/lib/component.js
+++ b/lib/component.js
@@ -125,6 +125,14 @@ export default class Component extends WebComponent {
     return true;
   }
 
+  toString() {
+    try {
+      return `${this.tagName}#${this.panelID}`;
+    } catch(e) {
+      return 'UNKNOWN COMPONENT';
+    }
+  }
+
   update(stateUpdate={}) {
     if (!this.initialized) {
       Object.assign(this.state, stateUpdate);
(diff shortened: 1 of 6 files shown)


lib/component.js
```diff
@@ -125,6 +128,14 @@ export default class Component extends WebComponent {
     return true;
   }
 
+  toString() {
+    try {
+      return `${this.tagName}#${this.panelID}`;
```
evnp: What error could we expect here?

test/test-component.js
```diff
@@ -256,3 +262,36 @@ describe('Nested Component instance', function() {
...
 });
+
+
+describe('Rendering exception', function() {
+  var el;
+
+  beforeEach(function(done) {
+    document.body.innerHTML = '';
+    el = document.createElement('breakable-app');
+    el.logError = sinon.spy();
```
evnp: What's this sinon?

test/test-component.js
```diff
@@ -256,3 +262,36 @@ describe('Nested Component instance', function() {
...
+    document.body.appendChild(el);
+    window.requestAnimationFrame(function() {
+      done();
+    });
+  });
+
+  it('does not prevent component from initializing', function() {
+    expect(el.initialized).to.be.ok;
```
evnp: ignore me if this is dumb, but the expect docs make it look like you want `to.be.ok()` https://github.com/Automattic/expect.js

Really find that a weird way to describe truthyness (hence having to look it up)... ah well
```
expect(everything).to.be.ok;
```

evnp:
Nice, I like it a lot % c

lib/component.js
tdumitrescu: In the current implementation, none within reason. But if for example `this.panelID` became a getter and executed code, it could potentially throw. `toString()` should be safe to call under basically any circumstances, especially considering that it's used for error output in a `catch` clause below. Few things more frustrating than getting an error which you can't even see because your error handler blew up.

test/test-component.js
tdumitrescu: http://sinonjs.org/ (bundled automatically by Web Component Tester)

test/test-component.js
tdumitrescu: Unfortunately we're not using `expect.js` here, but the lesser [Chai](http://chaijs.com/api/bdd/) (because wct automatically bundles it globally)

PR #12, opened 2016-07-23
added a function to return a parent reference by name.

chzhcn:


lib/component.js
```diff
@@ -161,6 +161,21 @@ class Component extends WebComponent {
...
   }
 
+  /**
+   * Returns the reference to the parent of the given name. An error is thrown out if no parent of
+   * such name is found.
+   * @param {tagName} - name of the parent to search for.  @example
+   * this.parentbyTagName('myApp');
+   */
+  parentByTagName(tagName) {
+    for (let node = this; node; node = node.parentNode) {
```
tdumitrescu: This looks like a generic DOM-traversal helper. Are you sure you're not looking for something to traverse the custom component hierarchy as your example seems to imply? i.e. `node = node.$panelParent;`

lib/component.js
```diff
@@ -161,6 +161,21 @@ class Component extends WebComponent {
...
+      if (node.tagName === tagName) {
```
tdumitrescu: i'd suggest lowercasing these so that callers don't have to remember to yell `parentByTagName('MY-APP')`

lib/component.js
```diff
@@ -161,6 +161,21 @@ class Component extends WebComponent {
...
+        return node;
+      }
+    }
+    throw `${tagName} not found`;
```
tdumitrescu: throw Errors, not strings

lib/component.js
```diff
@@ -161,6 +161,21 @@ class Component extends WebComponent {
     }
   }
 
+  /**
+   * Returns the reference to the parent of the given name. An error is thrown out if no parent of
+   * such name is found.
+   * @param {tagName} - name of the parent to search for.  @example
+   * this.parentbyTagName('myApp');
```
tdumitrescu: as far as examples go, this one will never work in practice; custom element names are required to have a dash in them, and browsers uppercase tag names, so this more correctly would be `this.parentbyTagName('MY-APP');` but see my previous comment about allowing `this.parentbyTagName('my-app');`

lib/component.js
```diff
@@ -161,6 +161,21 @@ class Component extends WebComponent {
     }
   }
 
+  /**
+   * Returns the reference to the parent of the given name. An error is thrown out if no parent of
+   * such name is found.
+   * @param {tagName} - name of the parent to search for.  @example
```
tdumitrescu: this should be `@param {string} tagName - bla bla`. see http://usejsdoc.org/tags-param.html
and put `@example` on its own line

tdumitrescu:
Sorry about the review spam. Can we also get a quick test case?

chzhcn:
addressing comments; will let you know when it's ready

chzhcn:
@tdumitrescu cool. PTAL; also a new release plz :)

lib/component.js
```diff
@@ -161,6 +161,21 @@ class Component extends WebComponent {
     }
   }
 
+  /**
+   * Returns the reference to the parent of the given name. An error is thrown out if no parent of
+   * such name is found.
+   * @param {string} tagName - name of the parent to search for.
+   * @example this.parentbyTagName('my-app');
+   */
+  parentByTagName(tagName) {
```
tdumitrescu: let's add `tagName = tagName.toLowerCase()` at the top of the method so the caller doesn't have to worry about case at all.

lib/component.js
```diff
@@ -161,6 +161,21 @@ class Component extends WebComponent {
     }
   }
 
+  /**
+   * Returns the reference to the parent of the given name. An error is thrown out if no parent of
+   * such name is found.
+   * @param {string} tagName - name of the parent to search for.
+   * @example this.parentbyTagName('my-app');
```
tdumitrescu: `parentByTagName`

tdumitrescu: and maybe a better name would be `findPanelParentByTagName` - it's a mouthful, but since panel components _are_ actual DOM elements, it's ambiguous otherwise whether you're talking about DOM parents or Panel parents.

lib/component.js
```diff
@@ -161,6 +161,21 @@ class Component extends WebComponent {
     }
   }
 
+  /**
+   * Returns the reference to the parent of the given name. An error is thrown out if no parent of
+   * such name is found.
+   * @param {string} tagName - name of the parent to search for.
```
tdumitrescu: tag name, not just name. this is an HTML concept

lib/component.js
```diff
@@ -200,13 +215,14 @@ class Component extends WebComponent {
         }
       }
       if (!this.$panelParent) {
-        throw `panel-parent ${parentID} not found`;
+        throw Error(`panel-parent ${parentID} not found`);
```
tdumitrescu: heh, thanks for fixing!

test/test-component.js
```diff
@@ -200,6 +200,19 @@ describe('Nested Component instance', function() {
       })
     })
 
+    it('successfully finds the panel root', function(done) {
```
tdumitrescu: It "successfully finds a panel parent node by a given tag name." We could make a separate `describe('findPanelParentByTagName')` block here, but maybe that's overkill at this stage.

test/test-component.js
```diff
@@ -200,6 +200,19 @@ describe('Nested Component instance', function() {
...
+      document.body.appendChild(el);
+      window.requestAnimationFrame(function(){
+        childEl = document.createElement('nested-child');
+        childEl.setAttribute('panel-parent', el.panelID);
+        el.appendChild(childEl);
+        window.requestAnimationFrame(function() {
+          expect(childEl.parentByTagName("nested-app")).to.equal(el);
```
tdumitrescu: single quotes (linter still needed)

lib/component.js
chzhcn: just double check: L172 should support both?

lib/component.js
chzhcn: re: `parentByTagName` comment: do you mean no `this` in example?

chzhcn: used `myWidget`

tdumitrescu: no, i was just pointing out that the capitalization in the doc didn't match the actual method name ('parentbyTagName' vs 'parentByTagName'), sorry about the confusion

lib/component.js
tdumitrescu: yeah, if you lowercase `tagName` at the start then the only comparison you need to do is `node.tagName.toLowerCase() === tagName` and the entire thing becomes totally case-insensitive.

tdumitrescu:
Published in v0.8.5, thanks!

chzhcn:
Thanks for the clean up!

PR #16, opened 2017-01-27
Remove domPatcher.el and reset initialized flag on detachedCallback

tdumitrescu:
opening for @jhp

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -264,6 +264,9 @@ class Component extends WebComponent {
 
   detachedCallback() {
     this.$panelParent && this.$panelParent.$panelChildren.delete(this);
+    this.el.removeChild(this.domPatcher.el);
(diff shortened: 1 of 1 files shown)


jbwyme:
may want to be a bit more defensive as well: https://www.dropbox.com/s/n2x1zwf9r48c527/Screenshot%202017-01-27%2015.05.54.png?dl=0

tdumitrescu:
How does that happen? Element never finishes initializing / never gets attached and yet somehow detachedCallback still fires?

jbwyme:
just delete a report from here: http://josh.dev.mixpanel.org/report/3/dashboard

tdumitrescu:
Published for expediency in 0.10.0-rc7, but let's still get these issues squared away.

jbwyme:
thanks!!

PR #18, opened 2017-05-01
Web Components v1

tdumitrescu:
Pretty straightforward, mostly just replacing polyfills and translating API 1-to-1. Safari 9 now works too as of webcomponent 1.1.0.


lib/component.js
```diff
@@ -275,6 +277,10 @@ class Component extends WebComponent {
     this.initialized = false;
   }
 
+  static get observedAttributes() {
```
jbwyme: I love that they added this, great for a little extra self documentation

examples/shadow-dom/README.md
```diff
@@ -0,0 +1,59 @@
...
+  // ...
+});
+```
+
+To install and run the example from this directory: `npm install && npm start`. The page will be served on `localhost:8080` by Webpack dev server.
+
+### Notes
+
+- Panel uses the Shadow DOM v1 API, which behaves similarly to the v0 API. See the description in Eric Bidelman's [Shadow DOM v1: Self-Contained Web Components](https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom).
+- In order to work in browsers which don't yet natively implement Shadow DOM v1 (such as Firefox as of April 2017), the example uses the [ShadyDOM polyfill](https://github.com/webcomponents/shadydom). NB since this polyfill is not yet available on NPM, the example installs it via Bower in an NPM `postinstall` script.
```
antialias: > since this polyfill is not yet available on NPM

looks like it is now: 
```sh
$ npm view @webcomponents/shadydom time.created
2017-05-11T19:26:25.342Z
```

tdumitrescu: 3223901064_88a288d103
thanks dude!

PR #20, opened 2017-06-23
removed babelrc from files published to npm

antialias:
not necessary to publish babelrc if it's getting built with babel


tdumitrescu:
thx 4 cleanup

PR #21, opened 2017-08-17
Panel devtools extension

nojvek:


nojvek:
Ping @tdumitrescu

tdumitrescu:
Woot will try it out in a bit!

nojvek:
@tdumitrescu Kind ping ^

scripts/publish-devtools.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * Usage: node scripts/publish-devtools.js
+ */
+
+const promisify = require(`promisify-node`);
+const fs = promisify(`fs`);
```
tdumitrescu: No biggie but are you actually using fs functions that don't have sync versions? e.g. it might simplify this a bit (and be more standard Node) just to do `fs.writeFileSync(...)` instead of `await fs.writeFile(...)`.

scripts/publish-devtools.js
```diff
@@ -0,0 +1,58 @@
...
+    console.log(`NOTE: Only users of mixpanel-chrome-extensions@googlegroups.com can publish this item.`);
+    console.log(`Paste following url in browser, authenticate and paste generate code\n`);
+    console.log(`${googleOauthUrl}/auth?response_type=code&scope=https://www.googleapis.com/auth/chromewebstore&client_id=${webstoreClientId}&redirect_uri=${redirectUri}`);
+    const oauthCode = readlineSync.question(`\ncode: `);
+    const api = new WebstoreApi(new TokenManager(oauthCode, webstoreClientId, webstoreClientSecret));
+
+    // Upload new item and publish
+    console.log(`\nUploading and publishing ...`);
+    await api.update(webstoreExtensionId, zipContents);
+    await api.publish(webstoreExtensionId);
```
tdumitrescu: Life is easy in 2017. This script would have been so gnarly 6 years ago...

.eslintrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "mixpanel",
+  "globals": {
+    "require": true,
```
tdumitrescu: Since we only use `require` for those few scripts in the devtools dir, I think it makes more sense to declare this in those files directly, to help enforce `import` consistency throughout the rest of the codebase.

scripts/publish-devtools.js
nojvek: Not really, I just made it a habit of using promisified versions of fs in node. In this case it doesn't matter since its a script and we don't care if it's blocking. Moved to *Sync versions

.eslintrc.json
nojvek: Fixed. 

Although I haven't fixed all the linting issues in the rest of code base as part of this PR. Didn't want the code changes in this PR to balloon out because of eslint issues. I just wanted to see the squiggles when editing the publish script.

I'll send you another PR for that.

tdumitrescu: Yeah, I've been holding off on standardizing formatting according to linter until after https://github.com/mixpanel/panel/pull/18 lands, just to avoid excessive merge cleanup...

PR #22, opened 2017-09-21
Adding ControlledComponent, StateController and StateStore

nojvek:
ControlledComponent explicitly initializes a controller that handles state management

Added tests for ControlledComponent


lib/state-store.js
```diff
@@ -0,0 +1,25 @@
...
+    this._state = {};
+  }
+
+  get state() {
+    return this._state;
+  }
+
+  update(props) {
+    Object.assign(this._state, props);
+    this._listeners.forEach(listener => listener());
```
andrewpurcell: This is a very coarse-grained subscription. Do you intend for subscribers to receive the message and then re-query their state through the controller API?

examples/controller/index.js
```diff
@@ -0,0 +1,42 @@
...
+    return this.state.count;
+  }
+}
+
+document.registerElement(`counter-app`, class extends ControlledComponent {
+  get config() {
+    return {
+      controller: new CounterController(),
+      template: () => h(`div.counter`, [
+        h(`div.val`, `Counter: ${this.controller.getCount()}`),
```
andrewpurcell: I like the notion of the StateController as a state-selector layer as well as state-mutator.

lib/controlled-component.js
```diff
@@ -0,0 +1,40 @@
...
+    this._unsubscribeUpdates = this.controller.subscribeUpdates(() => this._update());
+  }
+
+  detachedCallback() {
+    super.detachedCallback();
+    this._unsubscribeUpdates();
+  }
+
+  attributeChangedCallback() {
+    // Do nothing, component should explicitly pass this to controller for an update
```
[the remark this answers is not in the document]
nojvek: Yeah. Super() ends up doing auto attr to state magic if special kind of named attributes are found. It also then calls this.update() which is no longer usable.

lib/state-store.js
nojvek: Yeah the idea is the view simply listens for an update and requeries the state as part of its vdom render.  The only use of this is the wire the view’s render. It’s intended to be a dumb callback.

lib/state-controller.js
```diff
@@ -0,0 +1,33 @@
+import StateStore from './state-store';
+
+/**
+ * A StateController's only duty is to control state.
```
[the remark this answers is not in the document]
nojvek: Yes. Super classes can take services as options if they want. The controller is essentially the orchestra conductor for an app/component.

lib/state-controller.js
```diff
@@ -0,0 +1,33 @@
...
+ * Subclasses extend this and expose methods that call the controller's _update method
+ * Controller will default create its own state store unless one is passed.
+ * StateController subcalsses should only accept the dependencies they need via constructor
```
evnp: typo subcalsses

lib/state-controller.js
```diff
@@ -0,0 +1,33 @@
...
+
+  get defaultState() {
+    return {};
+  }
+
+  get state() {
+    return this._store.state;
+  }
+
+  // Discourage external users from using state directly
```
evnp: :heart:

lib/state-controller.js
```diff
@@ -0,0 +1,33 @@
...
+  _update(props) {
+    this._store.update(props);
+  }
+
+  subscribeUpdates(listener) {
```
evnp: I personally think this should also be "private" even though it's accessed in ControlledComponent base class. I believe we want to discourage subclasses of ControlledComponent from using this directly

lib/state-store.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  subscribeUpdates(listener) {
+    this._listeners.push(listener);
+    return () => {
+      this._listeners = this._listeners.filter(l => l !== listener);
+    };
```
evnp: Personally think a matching `unsubscribeUpdates` method would be a lot more clear than returning this unsubscribe here and storing it in ControlledComponent. That method could take `listener` as an arg, so you'd store a reference to that function in order to unsubscribe later

lib/controlled-component.js
```diff
@@ -17,24 +17,13 @@ export default class ControlledComponent extends Component {
...
     // Do nothing, component should explicitly pass this to controller for an update
     // Super class calls this.update() which will throw an error
   }
 
-  // Don't render if state hasn't changed since controlled components return immutable state
-  shouldUpdate(newState) {
-    if (this._lastState !== newState) {
-      this._lastState = newState;
-      return true;
-    }
```
evnp: What would be cool is if StateControllers could be responsible for a subtree of state instead of the whole thing, and components could do something like
```
shouldUpdate(newState) {
  this.controller.shouldUpdate(newState);
}
```
where the controller makes a check like you have above and only updates if the relevant subtree has changed. This would let us not re-render every single component on every single update - similar to the "pure component" style of building a react app, something I'm a big fan of and have wanted to emulate in panel. imo this is one of the big benefits of keeping all state changes immutable since you can then do this with constant-time state obj comparisons instead of deeply comparing state

lib/state-store.js
```diff
@@ -12,10 +12,13 @@ export default class StateStore {
   }
 
   update(props) {
-    Object.assign(this._state, props);
-    this._listeners.forEach(listener => listener());
+    // Always create a new state object.
+    // if lastState === newState then state hasn't changed
+    this._state = Object.assign({}, this._state, props);
```
evnp: :+1:

lib/controlled-component.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+    // Don't allow component's update directly
+    this._update = this.update;
+    this.update = () => { throw new Error(`update() not allowed from component. Use controller`); };
+    this._unsubscribeUpdates = this.controller.subscribeUpdates(() => this._update());
+  }
+
+  detachedCallback() {
+    super.detachedCallback();
+    this._unsubscribeUpdates();
```
evnp: Is this really so much better than
```
this.updateListener = () => this._update();
this.controller.subscribeUpdates(this.updateListener);
...
this.controller.unsubscribeUpdates(this.updateListener);
```
Can't help it, sorry :) I guess it's minor, but this just seems clearer to me, esp if this is a publicly exposed mechanism

PR #23, opened 2017-09-26
Webpack test/browser for sauce tests.

nojvek:
Before only fixtures were transpiled, not the tests themselves. This fixes it.

Context: Yesterday I was in pain with wct sauce tests failing only safari 9. Turns out the RC was we never correctly transpiled all our test files. 

NOTE: This shouldn't require a panel publish as it's a test only change.


test/browser/component-tests.js
```diff
@@ -2,170 +2,170 @@
 /* global sinon, expect */
 /* eslint no-unused-expressions:0 */
 
-describe(`Simple Component instance`, function() {
-  var el;
+describe(`Simple Component instance`, () => {
```
jbwyme: there's a reason not to use arrow functions in mocha tests. `this` is often uses to share state across tests and can break some things.

jbwyme: https://mochajs.org/#arrow-functions

nojvek:
We weren't using this so moved to arrow, but I see your point, if someone else in future wants to use this, we don't want to break the pattern. Moved back to function() for mocha related functions

PR #26, opened 2018-01-09
Adding panel hot loaders to reload jade, styl, controller without page refresh

nojvek:


hot/update-panel-elems.js
```diff
@@ -0,0 +1,22 @@
+module.exports = function updatePanelElems(elemName, updateFn) {
+  const elems = document.querySelectorAll(`body /deep/ ${elemName}`);
```
tdumitrescu: `/deep/` has been deprecated for a while now and I think this will break entirely with the upgrade in https://github.com/mixpanel/panel/pull/18. Not sure but https://developers.google.com/web/fundamentals/web-components/shadowdom#findall may be the only path forward here ATM.

hot/update-panel-elems.js
```diff
@@ -0,0 +1,22 @@
...
+  let numUpdated = 0;
+  for (const elem of elems) {
+    if (elem._config && elem.update) {
```
tdumitrescu: Is this intended to detect whether it's a panel component? I'd suggest that checking for the presence of `elem.panelID` is more robust (less prone to collisions and doesn't depend on internal prop).

hot/loader-helpers.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-env commonjs */
+const path = require(`path`);
+
+// Retreive elemName for hot injection from path convention
```
tdumitrescu: nit 'retrieve'

hot/loader-helpers.js
```diff
@@ -0,0 +1,22 @@
...
+//
+// elem name patterns look like this
+//    ./src/.../${elemName}/index.jade
+// OR ./src/.../${elemName}.jade
```
tdumitrescu: Interesting. Not a blocker for merging but at some point this could use some more complete explanation here of how this convention works, i.e. that module file name is expected to match HTML element tag name (so things like multiple element classes in a single file just won't work, right?).

hot/style-loader.js
```diff
@@ -0,0 +1,43 @@
...
+  if (!helpers.isDevServerHot(this.options)) {
+    return;
+  }
+
+  const moduleId = loaderUtils.stringifyRequest(this, `!!${remainingReq}`);
+  const loaderOpts = loaderUtils.getOptions(this) || {};
+  let moduleSource = ``;
+  let isShadowCss = loaderOpts.shadow || /\b(inline|shadow)\b/.test(this.resourceQuery);
+
+  if (isShadowCss) {
```
tdumitrescu: I think you could probably do this more generically/robustly at runtime by having your generated code look at `elem.getConfig('useShadowDom')` and/or `elem.getConfig('css')`.

hot/loader-helpers.js
nojvek: That's right.

Will add a comment

hot/style-loader.js
nojvek: I like that approach

hot/update-panel-elems.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-env commonjs */
+
+// 'That's not enough, we need to go deeper'
+// NOTE: document.querySelectorAll(`body /deep/ ${elemName}`) is deprecated, so we use recursion
+function findPanelElemsByName(rootElem, elemName) {
```
tdumitrescu: `findElementsByTagName`?

hot/loader-helpers.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-env commonjs */
+const path = require(`path`);
+
+// Retrieve elemName for hot injection from path convention
+//
+// elem name patterns look like this
+//    ./src/.../${elemName}/index.jade
+// OR ./src/.../${elemName}.jade
```
tdumitrescu: Another off the cuff thought for a future revision... Our naming convention works well when the different component subparts (template, style, 'controller') are distinguished by file type/extension, but it breaks down otherwise. For instance, back in the earliest days we used Jade for most templates but wrote some of the simpler ones directly with hyperscript in JS (`h('div')` etc.). A scheme that's resilient to using any file type/tooling could include info in the filename too, so that rather than `index.jade` it'd be `template.jade` or `template.js` or `template.hbs` or whatever. So a component dir might look like:
```
my-widget/
  index.js  // imports component class and registers custom element
  component.js  // defines and exports the component class but doesn't register it
  [controller.js]
  styles.*
  template.*
```

nojvek:
Will special case for for ‘template’ and ‘styles’. Thanks for the tip.
-- 

Thanks,
Noj.

nojvek:
+ panel@0.13.0

PR #28, opened 2018-01-24
fix shouldUpdate() getting state update instead of full state

tdumitrescu:
Longer term the right thing to do is probably to split out the existing method into `shouldUpdate()` (given the update fragment to apply) and `shouldRender()` (given the entire current state), but this should fix uses of the existing API broken by https://github.com/mixpanel/panel/pull/25/files#diff-265939b2d7ffc46f04bdad9c44d75688R387

(also adds infra for async/await in browser tests, though doesn't convert most existing tests)

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -383,9 +383,13 @@ class Component extends WebComponent {
   // a full state update to one component while sending only "shared" state
   // updates to the app root.
   updateSelfAndChildren(stateUpdate, options={}) {
+    if (!this.initialized) {
+      return;
+    }
+
     const {store, cascade} = options;
-    if (this.initialized && this.shouldUpdate(stateUpdate)) {
-      Object.assign(this[store], stateUpdate);
+    Object.assign(this[store], stateUpdate);
+    if (store !== `state` || this.shouldUpdate(this[store])) {
       this.domPatcher.update(this.state);
 
       if (cascade) {
--- a/test/browser/component.js
+++ b/test/browser/component.js
@@ -2,6 +2,9 @@
 /* global sinon, expect */
 /* eslint no-unused-expressions:0 */
 
+const nextAnimationFrame = () => new Promise(requestAnimationFrame);
+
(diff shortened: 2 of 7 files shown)


lib/component.js
```diff
@@ -383,9 +383,13 @@ class Component extends WebComponent {
   // a full state update to one component while sending only "shared" state
   // updates to the app root.
   updateSelfAndChildren(stateUpdate, options={}) {
+    if (!this.initialized) {
+      return;
+    }
+
     const {store, cascade} = options;
```
nojvek: Platform apps have their own version of mixpanel-common which has its own version of panel, so things blow up when options is an empty object.

Should we atleast default the store and cascade if they are not present?

```
const {store = `state`, cascade = true} = options
```

```
const stateStore = this[store];
Object.assign(stateStore, stateUpdate);
if (store !== `state` || this.shouldUpdate(stateStore)) {
```

wdyt ?

test/browser/component.js
```diff
@@ -2,6 +2,9 @@
 /* global sinon, expect */
 /* eslint no-unused-expressions:0 */
 
+const nextAnimationFrame = () => new Promise(requestAnimationFrame);
```
nojvek: Love it!

lib/component.js
tdumitrescu: I kind of think that this lib code shouldn't be defensive against a poor integration (multiple incompatible versions of the lib living in the same component hierarchy). The right thing is to be sure that all linked components are running the same code. In practice with the `mixpanel-common` thing, the simplest solution would be to export `panel` from within `mixpanel-common` so that an app can inherit from its classes or make its own components and be sure that there's only a single version of the Panel code. This is the same concept as how Panel exports the vdom `h()` function so that you don't have to worry about the installation/versioning of the virtual dom lib.

nojvek: Works for me

PR #30, opened 2018-03-14
Add pre- and post-update hooks

tdumitrescu:
It used to be that you could simply override a Component's `update()` method to wrap update behavior (for instance, for automatically persisting state to localStorage), but as of https://github.com/mixpanel/panel/pull/25 that no longer works with cascading updates, because the main `update()` method only runs on the component making the initial update call, while any other linked components only get updated via internal methods.

This branch addresses the issue by offering callback hooks `preUpdate` and `postUpdate`:
```js
get config() {
  return {
    // ...
    hooks: {
      postUpdate: () => persistLatestStuff(this.state),
    },
  };
}
```

We could alternatively make sure that cascading updates actually call `update()` on each affected component in the tree (which would be a bigger refactor), but IMO that's not really obvious behavior, and this is a little more explicit.

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -361,6 +361,14 @@ class Component extends WebComponent {
     return this._rendered || EMPTY_DIV;
   }
 
+  // run a user-defined hook with the given params, if configured
+  _runHook(hookName, ...params) {
+    const hook = (this.getConfig(`hooks`) || {})[hookName];
+    if (hook) {
+      hook(...params);
+    }
+  }
+
   _stateFromAttributes() {
     let state = {};
 
@@ -411,10 +419,7 @@ class Component extends WebComponent {
       return;
     }
 
-    const {preUpdate, postUpdate} = this.getConfig(`hooks`) || {};
-    if (preUpdate) {
-      preUpdate();
-    }
+    this._runHook(`preUpdate`, stateUpdate);
 
     const {store, cascade} = options;
     Object.assign(this[store], stateUpdate);
@@ -430,9 +435,7 @@ class Component extends WebComponent {
       }
     }
 
-    if (postUpdate) {
-      postUpdate();
-    }
+    this._runHook(`postUpdate`, stateUpdate);
   }
 }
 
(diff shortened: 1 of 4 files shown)


nojvek:
I really like the idea of not having to override `update()` in subclasses.

I'm not 100% convinced on hooks as the api. I understand snabbdom has this cool post/pre and other hooks but I see panel as a very light interface over webcomponents api that lets you bind state and function based renderer based on that state.

image

so if we follow the lifecycle pattern, it would make sense to simply have a `updatedCallback(stateUpdate)`.  This is where one would add logging, persisting of state e.t.c 

What do you think?

nojvek:
I’m not opposed to pre and post. Could someone explain when you’d want to use pre as a concrete example?

I’m simply advocating for simplicity.

tdumitrescu:
I absolutely share the ideal of keeping Panel as lightweight a connector as possible between the Web Components API and the Virtual DOM rendering layer.

If there are questions around naming, I am not at all wedded to the terminology of hooks vs callbacks or a host of alternatives (and I can't say that Snabbdom was the inspiration for this particular feature, I just wanted to solve the existing problem of how to run extra code on each state update).

That being said, one thing I explicitly don't want to do is make the Panel part of the API easily confusable with the standard Web Components API, which I think the `updatedCallback()` proposal is susceptible to. As of now, the Panel-specific parts are purposely largely confined to the `config` object - which I wouldn't mind renaming to `panelConfig` either, to make it more explicit.

tdumitrescu:
One thing a `preUpdate` hook could usefully do is replace the current `shouldUpdate()` method; return `false` to prevent the current component updating.

tdumitrescu:
(And FWIW I think update hooks are an easily abusable feature and would love to be able to avoid them altogether, but I don't think it's possible without hobbling useful functionality like state-change-logging and auto-persistence.)

tdumitrescu:
@jordanmnunez good idea, I added `stateUpdate` as a param, but held off on passing `this.state` for now until we see a real use case for hooks that don't already have access to the component's context.

PR #31, opened 2018-03-28
Proposal: ProxyComponent

andrewpurcell:
Putting this up to get some comments. I'm working on tests and generic examples.

WHY
For a project we're working on, we are evolving some existing components and would like to test the new ones out in beta. The safest way to roll out a -v2 behind a flag is to build a new component. Luckily, both versions have the same interface, defined as incoming attributes and outgoing events.

We don't currently have a concept in our architecture for separating component interface from implementation.

WHAT IS IT
ProxyComponent is an abstract base class for components which is designed to allow us to switch implementations seamlessly. It copies attributes from its own instance to that of its child, and re-dispatches events from its children to its parents.

EXPECTED USES
The first intended use is to roll out a new version of one of our widgets. It could also be used to implement A/B testing or feature flagging.


tdumitrescu:
I imagine you're putting this up in its current state as RFC - I think it would help the evaluation a whole lot to see an example or two. Maybe in https://github.com/mixpanel/panel/tree/b140b4293cb26eec0893a106da94dbcf3cb07df5/examples

andrewpurcell:
Did that example help? I can't get it to build and run locally, I copied most of the controller example config. The main problem I'm trying to solve with this is versioning components.

tdumitrescu:
I'll try to take a look this week and get the example running for you, that should help evaluation

tdumitrescu:
OK I fixed up the example and a few other issues. It seems like a fairly lightweight optional wrapper, which I like - I could see pulling this and `ControlledComponent` and related modules into a `component-utils` dir or something like that.

My main questions are around the benefits of this approach over a simple dynamic tag. AFAICT the main job of `ProxyComponent` is to choose a tag name to insert into the DOM, and the event-management stuff is just a necessary consequence of the wrapping. A no-wrapper version of the example template would look like:
```js
h(window.showTheNewCard ? `address-card-v2` : `address-card-v1`, {
  attrs: {name: `Ben`, city: `San Francisco`},
  on: {clickedCity: () => alert(`it's true!`)},
})
```
or in a jade version:
```jade
#{window.showTheNewCard ? `address-card-v2` : `address-card-v1`}(
  attrs={name: `Ben`, city: `San Francisco`}
  on={clickedCity: () => alert(`it's true!`)}
)
```
In these you could throw away the `address-card` proxy altogether (and in non-toy code probably pull the tag-choosing logic outside the template). Am I missing something critical there?

andrewpurcell:
Thanks for fixing it and thinking about it. I don't think the toy example is the best demonstration, which is why I attached the other PR in -common.

I think that the core question is: if we have some higher order component logic to decide which version of a component gets shown, where does that logic belong? To answer that question, we have to know:

* for a specific use case, how many times does our code have to make the decision? is it scoped to one or two places, or is it scattered across different modules and repositories?
* is there a real boundary between interface and implementation? i.e. do different people own the caller and receiver sides?

In the intended use case, see the linked PR, I will admit that the answers are: we make the decision twice and the team that works on this owns the parent and child components. This is a future looking move.

As for dynamic tag function versus a ProxyComponent subclass:

### Dynamic tag function
➕works right now
➕limited number of callers that have to compute which tag to display
➖non-standard way of rendering a component in jade (`=h()` or `#{function()}` versus `my-tag`)
➖easy to forget to use

### ProxyComponent subclass
➕can provide a semantically meaningful tag name
➕hides the decision making from the caller, no coupling to a specific function, only to a global tag name
➕easy to read in the parent component when you don't care about the specific subcomponent impl
➕provides a surface to evolve the interaction between the parent and child
* natural place to implement a decorator pattern
* can act as an adapter to shape and filters events and attributes
* can implement shared concerns like logging or tracking for both components

➖it is not the simplest solution that could work

Overall, I think that ProxyComponent represents a reasonable first step at implementing higher order components in our architecture. I think that HOCs will allow us to build simpler, better-scoped and more reusable components. We don't strictly need it for our current project, but I think it's a low-risk experiment.

tdumitrescu:
Thanks for the clear writeup. The use case of versioning and testing updates to an existing component that's used in a bunch of places is a convincing one; certainly better not to make multiple callers do switching logic. Let's get this cleaned up (TODOs/documentation/example readme) and ready to go!

tdumitrescu:
BTW not particularly substantive to the discussion, but tagname interpolation via `#{}` is perfectly standard jade/pug, and there are tests for it; see e.g. the fixture at https://github.com/pugjs/pug/blob/56d803caaa8a93a040b7be0b8a36abdc4ce8c509/packages/pug/test/cases/tag.interpolation.pug

andrewpurcell:
This should be much more complete now. I had a problem with the docs and all my new jsdoc gets dumped under `global`, but otherwise this is tested and exampled. LMK what you think!

andrewpurcell:
those look good to me. will you merge and publish?

tdumitrescu:
Published in v1.3.0

PR #36, opened 2018-06-13
use await nextAnimationFrame() in tests

### LINKED ISSUE #24: Update browser/component-tests to use `await raf()`
Currently we use `done()` and callbacks but async would be better style

```
      window.requestAnimationFrame(() => {
        childEl = el.getElementsByTagName(`nested-child`)[0];
        done();
      });
```

Example: test/server/component.js
```
    it('applies only the last state in a given frame', async function() {
       ...
      expect(domPatcher.el.textContent).to.eql('Value of foo: bar');
      await raf();
      expect(domPatcher.el.textContent).to.eql('Value of foo: whew');
    });
```

nojvek:
Fixes #24 

Use https://github.com/mixpanel/panel/pull/36/files?utf8=%E2%9C%93&diff=unified&w=1, makes reviewing a lot easier.

Tests passing 🐇 👻 🍏

### DIFF
--- a/test/browser/component-utils/controlled-app.js
+++ b/test/browser/component-utils/controlled-app.js
@@ -1,21 +1,25 @@
 /* eslint-env mocha */
 /* global expect */
 
+import requestAnimationFrameCB from 'raf';
+
+const raf = () => new Promise(requestAnimationFrameCB);
+
 describe(`Controlled App`, function() {
   let el;
 
-  beforeEach(function(done) {
+  beforeEach(async function() {
     document.body.innerHTML = ``;
     el = document.createElement(`controlled-app`);
     document.body.appendChild(el);
-    window.requestAnimationFrame(() => done());
+    await raf();
   });
 
   it(`does not allow update on component`, function() {
     expect(() => el.update({foo: `not bar`})).to.throw(/update\(\) not allowed from component. Use controller/);
   });
 
-  it(`Behaves like normal component`, function(done) {
+  it(`Behaves like normal component`, async function() {
     let count = 0;
     expect(el.controller.state).to.be.eql({count});
     expect(el.textContent).to.contain(`Counter: ${count}`);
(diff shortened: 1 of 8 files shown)


test/browser/component-utils/controlled-app.js
```diff
@@ -1,32 +1,33 @@
 /* eslint-env mocha */
 /* global expect */
 
+import requestAnimationFrameCB from 'raf';
+
+const raf = () => new Promise(requestAnimationFrameCB);
```
tdumitrescu: Rather than repeating this boilerplate in each browser test module, how about pulling it into a shared test util?
```js
import {nextAnimationFrame} from '../../util';
```
(FWIW I also prefer the existing usage of `await nextAnimationFrame()` to `await raf()`)

test/browser/component-utils/controlled-app.js
```diff
@@ -1,32 +1,33 @@
 /* eslint-env mocha */
 /* global expect */
 
+import requestAnimationFrameCB from 'raf';
```
mduan: What's the reason for using a polyfill rather than the browser's `requestAnimationFrame`?

nojvek: raf handles both node and browser. It also does polyfilling for different browsers like mozRequestAnimationFrame. I think now most browsers are consistent so may be we don't need it.

test/browser/component-utils/controlled-app.js
nojvek: Sure

test/browser/component-utils/controlled-app.js
tdumitrescu: Yeah I don't think we need it here. These are browser-only tests (Node tests are in `test/server/`), and every browser we support/test on has a working `requestAnimationFrame` implementation (as seen by the liberal use of it in the existing browser tests).

PR #39, opened 2018-08-20
Ability for Panel components to auto parse attributes -> this.attrs

nojvek:
- [x] To camel or not to camel 🐪 (decision: no camel)
 - [x] Tests & Fixtures
 - [x] Documentation


andrewpurcell:
Thanks for the executable proposal. I'd like to marinate on this for a couple days. I'm still not sold on the mirroring of attributes to state, although I'm very pro defining a schema and not worrying about s10n and where exactly my inputs come from (state vs attrs vs ??).

nojvek:
I understand we’ve had on and off discussions about this. My reasoning is the following. 

In Front ends, managing state is probably the biggest issue. Compared to jquery old school style where state is held in dom and required a ton of data- attr mangling, panel is a big improvement. 

However child panel components in CS terms have two stores of state. The attribute store and it’s state store. The attribute store requires a parse from string everytime you want to load a non string value. This means the template itself isn’t a function of state, but a function of state + attributes. 

With this, the mental model is simplified. A template is a function of state. Attrs are simply an  abstraction and require no manual parsing in templates or helpers. The store is always kept up to date. 

A number of panel components don’t update themselves when the attributes change. Mostly because they don’t hook into attrChangedHandler. 

Once you define the scheme, the component author doesn’t need to write boilerplate code. We can focus more on creating kickass self documented components, rather than fighting the nitty gritties of webcomponents.

tdumitrescu:
We've discussed this a bunch in other venues over the past year, so I'll try not to rehash old arguments too much.

Passing HTML attributes through an automatic and standardized parsing+validation stage would be useful - regardless of whether we end up autogenerating documentation from jsdoc or purely from component code (which I think comes with a lot of problems, mainly that it discourages descriptive docs in favor of a raw schema, unless we end up rolling our own crappy docs-in-code format).

I don't think reflecting HTML attrs directly into `state` is the best approach though. It opens up namespacing issues (e.g. accidentally shadowing an attribute with a separate state entry), and also works poorly with our current naming conventions + templating formats (e.g. dashed HTML attr `foo-bar` gets injected into a Jade template as...an illegal JS identifier?). More importantly, as @andrewpurcell mentioned, it flattens the distinct spaces of the component's internal state and its external API as represented in HTML attrs.

An API that I'd like for this might look something like this in an example usage:
```js
class MyWidget extends Component {
  static get htmlAttributes {
    return {
      'foo-bar': String,
      'baz': Number,
      'quux': JSON,
    };
  }
}
```
```jade
.my-widget
  .baz
    | I rerender every time HTML attribute 'baz' gets set
    | attr baz + 1 = #{$attrs.baz + 1}
```
I'd probably like to keep a parsed-attrs store entirely separate from `state`, have it updated every time an attr changes, and injected automatically into templates the way `$helpers` etc are currently injected. Also keeps the state-cascading code simpler, since attrs should always be specific to one component and not cascade to children.

nojvek:
Thanks @andrewpurcell and @tdumitrescu. Fair arguments from both of you. 

I think there are two things. 
1) have some form of attribute schema that auto-parses string attributes to proper js primitives. I like $attrs idea. I did sleep on it and came to similar conclusion.
2) have some form of documenting other things likes events and component usage. Fair that it doesn't belong in panel. I'll move it out of here. Keep this PR focused on attrs.

So here's my api proposal

```
static get attrsSchema() {
  return { 
    'foo-bar': { type: 'string'},
    'open': { type: 'boolean', propName: 'isOpen'},
    'sections': { type: 'json'},
    'blah-blah-blah': {type: 'number', propName: 'blah3x', description: 'this is blah attr. for humans'}
  }
}
```

Keeping it as an object so we can also provide custom prop names for attributes. By default it will convert dash-case to snakeCase.

Are you cool with it?

lib/component.js
```diff
@@ -396,6 +412,34 @@ class Component extends WebComponent {
...
   }
 
+  _updateAttrs(attr) {
+    const attrsSchema = this.constructor.attrsSchema;
+    if (attrsSchema.hasOwnProperty(attr)) {
+      const propName = attrsSchema[attr].propName || this._toCamelCase(attr);
+      const propType = attrsSchema[attr].type || `string`;
+      let propValue = null;
+
+      if (propType === `string`) {
```
tdumitrescu: I don't feel super strongly about this, but do you have a reason for preferring string type names to the [Vue-style types](https://vuejs.org/v2/guide/components-props.html#Prop-Types) @zakj and I have suggested in the past where this would be `if (propType === String)` etc?

test/server/component.js
```diff
@@ -113,17 +113,20 @@ describe(`Server-side component renderer`, function() {
...
     expect(html.toLowerCase()).to.contain(`<div class="attr-app">`);
     expect(html).to.contain(`Value of attribute wombats: 15`);
+    expect(el.attrs.wombats).to.equal(15);
   });
 
   it(`reacts to attribute updates`, async function() {
-    const el = new AttrReflectionApp();
+    const el = new AttrApp();
     el.setAttribute(`wombats`, `15`);
+    expect(el.attrs.wombats).to.equal(15);
```
tdumitrescu: 👍 when we go forward with this PR, let's ensure there are some more thorough test cases for `attrsSchema`, for instance testing every type, as well as behavior for invalid values.

lib/component.js
nojvek: I personally like strings because the entire attrSchema return can be jsoni-fied. This means I could create a component called `<attr-schema-doc component-name='hello world' attrs-schema="{schema json here}">`.

test/server/component.js
nojvek: oh yes. Most definitely. Created a todo list on the PR.

lib/component.js
```diff
@@ -396,6 +412,34 @@ class Component extends WebComponent {
     return state;
   }
 
+  _updateAttrs(attr) {
+    const attrsSchema = this.constructor.attrsSchema;
+    if (attrsSchema.hasOwnProperty(attr)) {
+      const propName = attrsSchema[attr].propName || this._toCamelCase(attr);
```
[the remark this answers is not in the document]
nojvek: Haha! One person's slippery slope is another person's water slide.

I personally prefer camel case because it reads better in template and has consistent access. i.e  suppose you have `progressive`, `sections`, `empty-message` and `search-filter`. I know its minor but I'm thinking as dev experience.

Option 1) They are accessed as `$attrs.progressive`, `$attrs.sections`, `$attrs['empty-message']`, `$attrs['search-filter]`.

Option 2) They are accessed as `$attrs.progressive`, `$attrs.sections`, `$attrs.emptyMessage`, `$attrs.searchFilter`.

Also dom element.style already give us a camel cased access so it's not a new thing. Plus with the auto camel casing, we have an enforced convention.

Let me poll others and see what they think.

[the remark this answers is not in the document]
tdumitrescu: Option 3 they are accessed as `$attrs['progressive']`, `$attrs['sections']`, `$attrs['empty-message']`, `$attrs['search-filter']` (which most closely matches the standard `getAttribute('myattr')` pattern which we already use)

lib/component.js
```diff
@@ -314,14 +323,32 @@ class Component extends WebComponent {
...
 
+  /**
+   * Attributes schema that defines the component's attributes and their types
+   * Panel auto parses attribute changes into this.attrs object and $attrs template helper
+   *
+   * @type {Object.<string, AttrSchema>}
+   */
+  static get attrsSchema() {
+    return {
+      'style-override': {
```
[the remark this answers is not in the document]
nojvek: style-override is going through a different path and IMHO doesn't make sense to be parsed into $attrs. https://github.com/mixpanel/panel/blob/master/lib/component.js#L323

Removing it from default attrsSchema too.

lib/component.js
```diff
@@ -15,6 +15,12 @@ const DOCUMENT_FRAGMENT_NODE = 11;
  *
  * @example <caption>Defining a Panel component</caption>
  * class MyWidget extends Component {
+ *   static get attrsSchema() {
```
tdumitrescu: I get the desire to document usage, but considering that this example is so bare-bones and doesn't even have `config` details, I'd prefer to leave it out of this example and document at the implementation point.

lib/component.js
```diff
@@ -213,6 +219,7 @@ class Component extends WebComponent {
     // appState and isStateShared of child components will be overwritten by parent/root
     // when the component is connected to the hierarchy
     this.state = {};
+    this.attrs= {};
```
tdumitrescu: I thought we had eslint hooked up here? Surprised it hasn't complained.

lib/component.js
```diff
@@ -396,6 +419,29 @@ class Component extends WebComponent {
     return state;
   }
 
+  _updateAttrs(attr) {
```
tdumitrescu: a brief comment description of this method would be useful (since its name sounds like it's performing multiple updates), also noting the side effect of updating `this.attrs` rather than returning anything

lib/index.d.ts
```diff
@@ -94,11 +108,31 @@ declare namespace Component {
...
+        /** Description of attribute, what it does e.t.c */
+        description?: string;
+    }
 }
 
 type ConfigOptions<State, AppState> = Component.ConfigOptions<State, AppState>;
 
 export class Component<State, AppState = {}> extends WebComponent {
+    /**
+     * Attributes schema that defines the component's attributes and their types
```
tdumitrescu: probably worth being explicit that it's HTML attributes in all this documentation

test/browser/component.js
```diff
@@ -1,9 +1,11 @@
 /* eslint-env mocha */
-/* global sinon, expect */
+/* global sinon, expect, chai */
 /* eslint no-unused-expressions:0 */
 
 import nextAnimationFrame from './nextAnimationFrame';
 
+chai.config.truncateThreshold = 0; // nicer deep equal errors
```
tdumitrescu: is there a better place for this than an individual test module?

test/browser/component.js
```diff
@@ -190,6 +192,84 @@ describe(`Simple Component instance`, function() {
...
+    expect(el.innerHTML).to.equal(trimHtml(`
+      <div class="attrs-reflection-app">
+        <p>str-attr: "hello world"</p>
+        <p>bool-attr: true</p>
+        <p>number-attr: 0</p>
+        <p>json-attr: null</p>
+      </div>
+    `));
+
+    expect(el.attrs).to.deep.equal({
```
tdumitrescu: This feels like a different test case than the `innerHTML` check above

test/fixtures/attrs-reflection-app.js
```diff
@@ -0,0 +1,21 @@
...
+
+export class AttrsReflectionApp extends Component {
+  static get attrsSchema() {
+    return {
+      'str-attr': {type: `string`},
+      'bool-attr': {type: `boolean`, default: true},
+      'number-attr': {type: `number`, default: 0},
+      'json-attr': {type: `json`},
+
+
```
tdumitrescu: spaaace

test/server/component.js
tdumitrescu: This branch has some decent high-level coverage now, but I still feel like I'd like to see at least one chunk of server-side tests that run through the attrs cases a little more systematically and focused rather than as part of bigger rendering tests. For instance with cases divided up by attr type, including both happy path and null/undefined/0/''/malformed-json cases, simply calling `setAttribute()` and checking what ends up in `attrs` without worrying about the rendering cycle.

lib/component.js
```diff
@@ -314,14 +323,27 @@ class Component extends WebComponent {
     this.initialized = false;
   }
 
+  /**
+   * Attributes schema that defines the component's attributes and their types
+   * Panel auto parses attribute changes into this.attrs object and $attrs template helper
+   *
+   * @type {Object.<string, AttrSchema>}
```
tdumitrescu: can't document the schema format here?

test/fixtures/attrs-reflection-app.js
nojvek: you want aligned key value pairs ? Not sure what you mean?

lib/component.js
nojvek: I did it, but it didn't show up in jsdoc :/ So I kept it simple. I'll put it back.

test/fixtures/attrs-reflection-app.js
tdumitrescu: i don't want these random two lines of whitespace

lib/component.js
nojvek: We still eslint everything. We've never enabled the `space-infix-ops` rule because AFAIK it's been an age old debate since the time I joined. 

We wanted to keep things consistent with python in function calls e.g `function(a=true, b='hello) {`. eslint doesn't give us a rule per operator so we didn't enabled it at the sacrifice of little things like this slipping in.

PR #40, opened 2018-08-31
Show state, component, appState and attrs in panel devtools

nojvek:
image

### DIFF
--- a/devtools/devtools.js
+++ b/devtools/devtools.js
@@ -1,5 +1,6 @@
 /* global chrome */
 // The function below is executed in the context of the inspected page.
+// Event listener global scope is slightly different to inspector scope, so we test for window[`$0`]
 function getPanelElementState() {
   if (!window[`__$panelDevToolsReady`] && window[`$0`] && window.document.body) {
     // Chrome extension api doesn't let us know when expression has been edited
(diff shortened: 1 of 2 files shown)


devtools/devtools.js
```diff
@@ -1,33 +1,56 @@
...
-    if (selectedElem.controller && selectedElem.controller.state) {
-      window[`__$panelDevToolsLastSelectedElem`] = selectedElem;
-      selectedElem.controller._update();
-      return selectedElem.controller.state;
+    // Go through ancestors to find a panel element. panelID is used as signifier for a panel component
+    // This is so user can right click -> inspect element anywhere inside a panel component and see its debug info
+    let panelElem = selectedElem;
+    while (!panelElem.panelID && (panelElem.parentElement || panelElem.getRootNode().host)) {
+      // using getRootNode().host to jump through shadow dom boundaries as well
+      panelElem = panelElem.parentElement || panelElem.getRootNode().host;
```
tdumitrescu: doesn't this loop need to exit if `panelElem` doesn't exist? i.e. when the selected element is not within a panel component

devtools/devtools.js
```diff
@@ -1,33 +1,56 @@
...
     }
-    else if (selectedElem.state && selectedElem.update) {
-      window[`__$panelDevToolsLastSelectedElem`] = selectedElem;
-      selectedElem.update();
-      return selectedElem.state;
+
+    if (panelElem.panelID) {
```
tdumitrescu: if my last comment is legit then this also needs to be `if (panelElem && panelElem.panelID)`

devtools/devtools.js
```diff
@@ -1,33 +1,56 @@
...
+      else if (panelElem.update) {
+        panelElem.update();
+        debugInfo.state = panelElem.state;
+      }
+
+      if (!panelElem.isStateShared && panelElem.appState) {
+        debugInfo.appState = panelElem.appState;
+      }
+
+      if (panelElem.attrsSchema && Object.keys(panelElem.attrsSchema).length) {
```
tdumitrescu: Did you make `attrsSchema` available as an instance prop/getter? I thought it was `static`

devtools/devtools.js
```diff
@@ -1,33 +1,56 @@
...
+        debugInfo.attrs = panelElem.attrs;
```
tdumitrescu: What would you think of adding schema info (type, description) to the entries in `debugInfo.attrs`? so attr `foo-bar=llamas` might be rendered as something like
```js
{
  'foo-bar': {value: "llamas", type: "string", description: "I like turtles"}
}
```
Most components shouldn't have so many attrs that it gets huge

nojvek: I think we should have debugInfo show the actual object, rather than a merged copy of those two objects. Easier for user to understand what's inside attrs as they pipe things in the template. I'm showing both attrsSchema and attrs now, so it should satisfy the requirement.

PR #41, opened 2018-09-05
Support attrSchema definitions as simple strings as well as objects

nojvek:


lib/component.js
```diff
@@ -427,19 +427,22 @@ class Component extends WebComponent {
     const attrsSchema = this.constructor.attrsSchema;
     if (attrsSchema.hasOwnProperty(attr)) {
       const attrSchema = attrsSchema[attr];
```
zakj: nit: I would just normalize it once here, so the rest of this block doesn't have to know about the shortcut. e.g.
```javascript
let attrSchema = attrsSchema[attr];
if (typeof attrSchema === `string`) {
  attrSchema = {type: attrSchema};
}
```

lib/component.js
```diff
@@ -425,13 +425,25 @@ class Component extends WebComponent {
    */
   _updateAttr(attr) {
     const attrsSchema = this.constructor.attrsSchema;
+    const typeDefaults = {string: ``, boolean: false, number: 0, json: null};
```
tdumitrescu: pull this constant out of the method rather than redefining it on every call

test/server/component.js
```diff
@@ -175,4 +176,11 @@ describe(`Server-side component renderer`, function() {
       </div>
     `));
   });
+
+  it(`throws error if there is a malformed attrsSchema type`, function() {
```
tdumitrescu: :+1: thanks for error case test

PR #43, opened 2018-09-10
_syncAttrs on constructor and connectedCallback

nojvek:
It is possible to access this.attrs before calling super.connectedCallback or sometimes in defaultState. This also adds a this._syncAttrs helper


tdumitrescu:
Behavior change seems fine; let's lock it down with a test please.

nojvek:
image

Seems like travis got it now :)

PR #45, opened 2018-09-28
Show constructor in panel devtools so we can do 'show function definition'

nojvek:


nojvek:
That's not a bad idea. I added the constructor so its easy to find the source of the component quite easily, but lifecycle methods would be nice too.

nojvek:
Soz for the late reply, I got quite a few things on my plate, but will try to get this out before EOW

nojvek:
This has been sitting for almost a month. I'm gonna merge this and do the connectedCallback info at a later date. Hopefully its cool with you.

PR #47, opened 2018-12-05
Make it possible to specify Attribute types

jbwyme:
Expose a generic for "Attributes" so we can do static analysis of Attribute types. 

Example:

```
/**
 * @typedef {Object} State
 * @property {boolean} isMenuOpen
 * @property {?Annotation} editing
 */

/**
* @typedef {Object} Attributes
* @property {Annotation[]} annotations
*/

/**
 * @extends {Component<State, Attributes>}
 */
```

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -63,12 +63,12 @@ declare namespace Component {
         [hookName: string]: (params: any) => void;
     }
 
-    interface TemplateScope<AppState = {}> {
+    interface TemplateScope<Attributes = {}, AppState = {}> {
         /** AppState of the root panel component */
         $app: AppState;
 
         /** Attributes parsed from component's html attributes using attrsSchema */
-        $attrs: {[attr: string]: any};
+        $attrs: Attributes;
 
         /** A reference to the component itself */
         $component: WebComponent;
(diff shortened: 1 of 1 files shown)


tdumitrescu:
Can you include usage in a test fixture? That will get checked on CI with `tsc` if you mark the module with `// @ts-check`. See https://github.com/mixpanel/panel/blob/c7e40e2226adb82c073ee7c91859248e1bfdf5ef/test/fixtures/simple-app.js for an example.

jbwyme:
It seems to me that `AppState` is likely the least used generic. How about `Component<App, State, Attributes, AppState>`?

jbwyme:
Actually, is it possible to infer `State` from `App`??

nojvek:
I agree with you on that. 

@tdumitrescu how would you feel if we removed this.appState property from panel completely, and instead made all it's users explicitly access it via `this.app.state` ?

same with `this.updateApp` to `this.app.update()` ?

Kind of brings simplicity to panel. wdyt ?

jbwyme:
> I agree with you on that.
> 
> @tdumitrescu how would you feel if we removed this.appState property from panel completely, and instead made all it's users explicitly access it via `this.app.state` ?
> 
> same with `this.updateApp` to `this.app.update()` ?
> 
> Kind of brings simplicity to panel. wdyt ?

I like that. How do you suggest we handle local state?

tdumitrescu:
Mentioned elsewhere, but I think my favorite API right now would be to have explicit `localState` and `appState` available on any component, and `updateLocal()` and `updateApp()` to manipulate local vs shared state. Feels nicer than adding more dependencies on `this.app`, which tends to invite abuse.

nojvek:
Closing this in favor of #62 . In additional to runtime attr presence checking, it allows user to specify attribute types so compile time checking is possible too.

PR #48, opened 2018-12-13
Updates HMR to be compatible with webpack 4

pablofierro:
- Updates HMR panel loaders to be compatible with webpack 4
- Allows a more flexible HMR updating strategy by allowing a dynamic resolve of the components tag name `(based on the module's path)`.
- Removes base style loaders and defaults to a loader in charge of hot replacement for components with `shadowDom`, the reason for this is to avoid maintaining yet another style loader but leverage on [`style-loader`](https://github.com/webpack-contrib/style-loader) or whatever other style loaders there are out there.


package.json
```diff
@@ -46,6 +46,7 @@
     "loader-utils": "1.1.0",
     "lodash.pick": "4.4.0",
     "raf": "3.2.0",
+    "schema-utils": "^1.0.0",
```
tdumitrescu: pin version plz

hot/update-panel-elems.js
```diff
@@ -25,8 +25,8 @@ module.exports = function updatePanelElems(elemName, updateFn) {
 
   for (const elem of elems) {
     if (updateFn.call(null, elem)) {
-      const update = elem._update || elem.update;
-      numUpdated += update.apply(elem) ? 1 : 0;
+      numUpdated += 1;
```
tdumitrescu: ++

pablofierro: 0702f3a

package.json
pablofierro: 0702f3a

tdumitrescu:
Published in v1.5.1, thanks @pablofierro

PR #49, opened 2018-12-26
Add `initialized` to type definition

jbwyme:


lib/index.d.ts
```diff
@@ -135,9 +135,15 @@ export class Component<State, AppState = {}> extends WebComponent {
     /** Attributes parsed from component's html attributes using attrsSchema */
     attrs: {[attr: string]: any};
 
+    /** A reference to the top-level component */
+    app: Component<State, AppState>;
```
jbwyme: I'm not sure how to handle the generics here. Ideally, we'd figure out a way for `this.app` to actually know about the correct component type.

tdumitrescu: Not sure what's going on in this file TBH, but any component can have both `state` and `appState` members simultaneously (that's the point of `appState`, to have both a local store and a shared store available). There's a prop `isStateShared` which is true when the basic default `state` store is shared (instead of a separate `appState` store).

PR #50, opened 2019-01-15
Improve template error reporting for Firefox

pablofierro:
Minor improvement on template error reporting:

After:
image

Before:
image

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -384,7 +384,7 @@ class Component extends WebComponent {
           $attrs: this.attrs,
         }));
       } catch (e) {
-        this.logError(`Error while rendering ${this.toString()}`, this, e.stack);
+        this.logError(`Error while rendering ${this.toString()}`, this, `\n`, e);
       }
     }
     return this._rendered || EMPTY_DIV;


lib/component.js
```diff
@@ -384,7 +384,7 @@ class Component extends WebComponent {
           $attrs: this.attrs,
         }));
       } catch (e) {
-        this.logError(`Error while rendering ${this.toString()}`, this, e.stack);
+        this.logError(`Error while rendering`, this, `\n`, e);
```
tdumitrescu: Can we please retain `this.toString()` so the element's panelID is still displayed?
```suggestion
        this.logError(`Error while rendering ${this.toString()}`, this, `\n`, e);
```

pablofierro: @tdumitrescu sure, reason I removed it is you see the actual DOM node the error occurred in, you can inspect it manually, is there a historical reason to print the element's `panelID` ?

tdumitrescu: It's just for easier debugging. I realize you can click through to the element and get a reference to it but I find it useful to have the ID readily available in the console, especially for situations like debugging errors within lists of similar elements.

PR #53, opened 2019-01-25
enum error messages

nojvek:


nojvek:
Yeah I didn't see a usecase for num enums yet and bools are true/false enums anyway. Json enums incur the perf hit of deepEquals so keeping it simple right now.

PR #57, opened 2019-03-02
dom shim customElements.get

nojvek:


lib/isorender/dom-shims.js
```diff
@@ -84,11 +84,14 @@ Document.prototype.createElement = function(tagName) {
   return el;
 };
 
-global.customElements = global.customElements || {
+global.customElements = {
```
nojvek: I'm writing tests for mp-table, and it seems that environment has a globals.customElements which isn't bound to registeredElements. This makes the customElements (get, define) bound to the registry as expected.

lib/isorender/dom-shims.js
```diff
@@ -84,11 +84,14 @@ Document.prototype.createElement = function(tagName) {
...
+  get(tagName) {
+    return registeredElements[tagName];
+  },
   define(tagName, proto) {
     tagName = tagName.toLowerCase();
     if (registeredElements[tagName]) {
-      throw DOMException(`Registration failed for type '${tagName}'. A type with that name is already registered.`);
+      throw new Error(`Registration failed for type '${tagName}'. A type with that name is already registered.`);
```
nojvek: Turns out DOMException wasn't a thing. Just throwing a regular error.

lib/isorender/dom-shims.js
```diff
@@ -84,11 +84,14 @@ Document.prototype.createElement = function(tagName) {
   return el;
 };
 
-global.customElements = global.customElements || {
```
tdumitrescu: No big deal but why remove capabilities? (in this case to define/shim `customElements` elsewhere before loading this module)

lib/isorender/dom-shims.js
```diff
@@ -84,11 +84,14 @@ Document.prototype.createElement = function(tagName) {
...
+global.customElements = {
+  get(tagName) {
+    return registeredElements[tagName];
+  },
```
tdumitrescu:
```suggestion
  },

```

lib/isorender/dom-shims.js
```diff
@@ -85,12 +85,16 @@ Document.prototype.createElement = function(tagName) {
...
+  get(tagName) {
+    return customElementsRegistry[tagName];
+  },
+
   define(tagName, proto) {
     tagName = tagName.toLowerCase();
-    if (registeredElements[tagName]) {
-      throw DOMException(`Registration failed for type '${tagName}'. A type with that name is already registered.`);
+    if (customElementsRegistry[tagName]) {
+      throw new Error(`CustomElementRegistryError: '${tagName}' name has already been registered`);
```
tdumitrescu: Out of curiosity, where does this message come from? The previous one came from Chrome's implementation at the time I think.. Current Chrome is
```
"Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry
    at <location>>"
```

lib/isorender/dom-shims.js
nojvek: Put it back, made the customElement registry not override itself. It seems mochapack is somehow initializing the module twice. So much state :(

test/server/dom-shims.js
```diff
@@ -0,0 +1,19 @@
...
+import '../../lib/isorender/dom-shims';
+
+import {expect} from 'chai';
+
+import {SimpleApp} from '../fixtures/simple-app';
+
+customElements.define(`my-app`, SimpleApp);
+
+describe(`customElements registry`, function() {
+  it(`.get() returns component class`, function() {
```
tdumitrescu:
```js
  describe(`.get()`, function() {
    it(`returns component class`, function() {
      // etc
    });
  });
```

lib/isorender/dom-shims.js
tdumitrescu: smh for mochapack

lib/isorender/dom-shims.js
nojvek: Nah just a rewording of the chrome error. Can make it exactly the same if you want.

test/server/dom-shims.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-env mocha */
```
tdumitrescu:
```suggestion
/* eslint-env mocha */

```

test/server/dom-shims.js
```diff
@@ -0,0 +1,22 @@
...
+import {expect} from 'chai';
+
+import {SimpleApp} from '../fixtures/simple-app';
+
+customElements.define(`my-app`, SimpleApp);
+
+describe(`customElements registry`, function() {
+  describe(`.get()`, function() {
+    it(`returns component class`, function() {
+      expect(customElements.get(`my-app`)).to.eql(SimpleApp);
```
tdumitrescu: strict equality should work here, right?
```suggestion
      expect(customElements.get(`my-app`)).to.equal(SimpleApp);
```

test/server/dom-shims.js
```diff
@@ -0,0 +1,22 @@
...
+    });
+  });
+
+  describe(`.define()`, function() {
+    it(`called twice throws an error`, function() {
```
tdumitrescu:
```suggestion
    it(`throws an error if called twice with the same tag name`, function() {
```

test/server/dom-shims.js
nojvek: oh is that what the difference between .eql and .equal was. I assumed .eql was an alias. #TIL

tdumitrescu: `eql` in chai is deep-equals, but afaik we can rely on comparing these by reference

PR #58, opened 2019-03-16
isorender support for components with useShadowDom: true

nojvek:


jbwyme:
what does this do?

nojvek:
@jbwyme nothing much. Just polyfills custom elements api better so we can mocha test panel components (with shadowDom:true) without having to invoke a big fat browser. 

Context: I like that panel can render on the server (nodejs env) with it’s isorender utils. I’m slowly trying to scratch an itch where we can test panel components with a fast test dev cycle with mocha.

tdumitrescu:
Yeah, I originally hacked up the SSR module with the idea of using it for faster, more deterministic functional component tests. Super stoked to see @nojvek filling in the shadow dom gap, which makes it much more broadly usable.

Testing aside, we'll be at a point soon where we can take advantage of SSR in Mixpanel's asset-serving infrastructure, and shadow dom support is a hard requirement there.

lib/isorender/dom-shims.js
```diff
@@ -61,6 +61,10 @@ class HTMLElement extends Element {
     }
     return this.__observedAttrs.includes(name);
   }
+
+  attachShadow() {
+    return document.createElement(`#shadow-root`);
```
tdumitrescu: Interesting - this works I guess because `html-element` doesn't check tag names strictly, but `#shadow-root` is an invalid tag name so this would break if the shim became more realistic (Chrome for instance throws an exception if you try this).

nojvek: I modelled it after what dev-tools showed me. Happy to change it to something else

image

nojvek: calling it `shadow-root`

test/server/dom-shims.js
```diff
@@ -21,3 +24,18 @@ describe(`customElements registry`, function() {
         .to.throw(`Registration failed for type 'my-app'. A type with that name is already registered.`); });
   });
 });
+
+describe(`customElement with shadowDom`, function() {
+  it(`el contains #shadow-root`, function() {
```
tdumitrescu:
```suggestion
  it(`el contains shadow-root`, function() {
```

PR #59, opened 2019-03-21
Router improvements

evnp:
1) Fixes a bug where hash is not decoded before comparison with fragment to decide whether update is necessary in `replaceHash`; this can result in circular hash update loop, see repro scenario.
2) Allows passing `{historyMethod: 'replaceState'}` to `replaceHash` so that you can make a singular  hash update from the client without adding a history entry if desired. My use case is lexicon, where currently nav to `report/123/lexicon` redirects to `report/123/lexicon#schema` (the hash is sanitized to match the default "subreport") and if the user now clicks back, they go back to `report/123/lexicon` instead of their previous report. This could be done by directly calling `window.history.replaceState` but doing through the panel router keeps code consistent and gives us the "has fragment changed?" check.

Repro for (1): check out `c7b2c515aad` in analytics, run lexicon in standalone mode. Nav to http://localhost:8088/iron/apps/lexicon/compiled/?feature-data_schema=true&project_id=4#transformations/events/[AI]%20Bookmark%20Retrieved. You should see
```
router.js:151 Uncaught RangeError: Maximum call stack size exceeded
    at router.js:126
    at Router.navigate (router.js:134)
    at Router.navigateToHash (router.js:66)
    at History.window.history.(:8088/iron/apps/lexicon/compiled/anonymous function) [as pushState] (http://localhost:8088/iron/apps/lexicon/compiled/lexicon.bundle.js:36858:15)
    at Router.replaceHash (router.js:163)
    at HTMLElement.index (index.js:118)
    at router.js:126
    at Router.navigate (router.js:134)
    at Router.navigateToHash (router.js:66)
    at History.window.history.(:8088/iron/apps/lexicon/compiled/anonymous function) [as pushState] (http://localhost:8088/iron/apps/lexicon/compiled/lexicon.bundle.js:36858:15)
```

Screen Shot 2019-03-20 at 6 54 21 PM

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -95,7 +95,8 @@ export default class Router {
     console.error(`No route found matching #${fragment}`);
   }
 
-  replaceHash(fragment) {
+  replaceHash(fragment, {historyMethod=null}={}) {
+    historyMethod = historyMethod || this.historyMethod;
     fragment = stripHash(fragment);
     const currHash = stripHash(this.window.location.hash);
 
@@ -103,7 +104,7 @@ export default class Router {
     // written to url, making `#bar baz` and `#bar%20baz` effectively the same.
     // This can result in hash update loops if the client passes in decoded hash.
     if (decodeURIComponent(fragment) !== decodeURIComponent(currHash)) {
-      this.window.history[this.historyMethod](null, null, `#${fragment}`);
+      this.window.history[historyMethod](null, null, `#${fragment}`);
     }
   }
 }
--- a/lib/router.js
+++ b/lib/router.js
@@ -97,7 +97,12 @@ export default class Router {
 
   replaceHash(fragment) {
     fragment = stripHash(fragment);
-    if (fragment !== stripHash(this.window.location.hash)) {
+    const currHash = stripHash(this.window.location.hash);
+
+    // decodeURIComponent since hash fragments are encoded while being
+    // written to url, making `#bar baz` and `#bar%20baz` effectively the same.
+    // This can result in hash update loops if the client passes in decoded hash.
+    if (decodeURIComponent(fragment) !== decodeURIComponent(currHash)) {
       this.window.history[this.historyMethod](null, null, `#${fragment}`);
     }
   }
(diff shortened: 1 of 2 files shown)


lib/router.js
```diff
@@ -95,10 +95,12 @@ export default class Router {
...
   }
 
-  replaceHash(fragment) {
+  replaceHash(fragment, {historyMethod=null}={}) {
+    historyMethod = historyMethod || this.historyMethod;
     fragment = stripHash(fragment);
-    if (fragment !== stripHash(this.window.location.hash)) {
-      this.window.history[this.historyMethod](null, null, `#${fragment}`);
+    const currHash = stripHash(this.window.location.hash);
+    if (decodeURIComponent(fragment) !== decodeURIComponent(currHash)) {
```
nojvek: may want to add a comment here why we can't just compare fragment and currHash. Why is decodeURIComponent important here.

test/browser/router.js
```diff
@@ -130,4 +130,49 @@ describe(`Router`, function() {
       expect(this.routerApp.textContent).to.equal(`Number: 42`);
     });
   });
+
+  describe.only(`replaceHash()`, function() {
```
tdumitrescu:
```suggestion
  describe(`replaceHash()`, function() {
```

evnp: oof, thanks

lib/router.js
evnp: added to a168ae4

tdumitrescu: Do we need to uri-decode anywhere else? like https://github.com/mixpanel/panel/blob/549ddd9db965485016e7ef3db76038b60a556681/lib/router.js#L61-L62

evnp:
Great question, I'll check it out

evnp:
@tdumitrescu after looking at the code for a bit, I think it does make sense to decode in `navigate` - we pass `window.location.hash` directly in as `fragment` [here](https://github.com/mixpanel/panel/blob/master/lib/router.js#L44) which could be compared to a fragment passed in manually (not window.location.hash) and likely not encoded. Tests pass, and this gets deployed to all iron apps at once now right? So I'll be able to do a complete QA pass after staging.

PR #60, opened 2019-03-28
render async on Promise rather than requestAnimationFrame

nojvek:
Render on promise completion rather than requestAnimationFrame. This keeps update async but avoids unnecessary layouts and paints

image

Single render at the end: 
image

This is a proof of concept


nojvek:
Closing it, raf is already good. It gives browser breathing room between renders to accept user input. May re-visit at later date if we really really need this.

PR #62, opened 2019-04-22
attr() function for validated attr access

nojvek:
We'll need to perform a refactor. 

In JS: Rather than accessing `this.attrs[attr]`, access it as `this.attr(attr)`.
In Jade: `$attrs[attr]` -> `$attr(attr)`

This gives us runtime validation if attributes not defined in attrSchema try to be accessed (if misspelt or removed). 

For deprecated attrs, adding a deprecatedMsg in console warns to developer that the attribute is deprecated.


nojvek:
This is just an idea. 

The alternative I am thinking is to wrap this.attrs in a proxy which does the get validation. With that soln no client refactor will be needed. 

It feels like a more elegant solution but i’ll need to figure out tests since deep equal object and proxy don’t quite work. 

Let me know the thoughts. 

Ideally I’m gonna give proxy a second shot and see if we can get a more elegant solution. 

The meat of validation logic will probably remain the same though. So feel free to review that.

nojvek:
Proposal #2: https://github.com/mixpanel/panel/pull/63 

I do like it better, since it will require no client refactoring changes

lib/component.js
```diff
@@ -505,10 +507,34 @@ class Component extends WebComponent {
         attrValue = this.getJSONAttribute(attr);
       }
 
-      this.attrs[attr] = attrValue;
+      this._attrs[attr] = attrValue;
+
+      if (attrSchema.deprecatedMsg) {
+        console.warn(`attr '${attr}' is deprecated. ${attrSchema.deprecatedMsg}`);
```
evnp: great idea

tdumitrescu: Would probably be useful to include the component name in the message

lib/component.js
```diff
@@ -505,10 +507,34 @@ class Component extends WebComponent {
...
 
+  /**
+   * gets the parsed value of an attribute
+   * @param {string} attr - attribute name
+   */
+  attr(attr) {
+    if (attr in this._attrs) {
+      return this._attrs[attr];
+    } else {
+      throw new TypeError(`attr '${attr}' is not defined in attrSchema`);
```
tdumitrescu:
```suggestion
      throw new TypeError(`attr '${attr}' is not defined in attrsSchema`);
```

lib/component.js
```diff
@@ -505,10 +507,34 @@ class Component extends WebComponent {
         attrValue = this.getJSONAttribute(attr);
       }
 
-      this.attrs[attr] = attrValue;
+      this._attrs[attr] = attrValue;
+
+      if (attrSchema.deprecatedMsg) {
+        console.warn(`${(this.nodeName || ``).toLowerCase()}: attr '${attr}' is deprecated. ${attrSchema.deprecatedMsg}`);
```
tdumitrescu: Component already implements `toString`, so this should work:
```suggestion
        console.warn(`${this}: attr '${attr}' is deprecated. ${attrSchema.deprecatedMsg}`);
```

nojvek: That kinda felt too noisy with `SIMPLE-APP#cjvogk9f800003h9hr8wdljcu`. What are your thoughts if we changed the .toString to be simply the lowerCase version of nodeName ?

tdumitrescu: I get the feeling you mainly care about lowercasing. Feel free to lowercase. I don't want to remove the instance ID, as that is actually very useful for debugging.

nojvek: I did mind the long panelID which is gibberish but not a biggie. used ${this} as you suggested.

nojvek:
Made the new stuff typescript friendly. Now we can validate attr access at compile time too. We'll need to figure out some way to auto-generate the TS interface. But it works

image

@jbwyme you may like this ^

PR #63, opened 2019-04-22
validate this.attrs[attr] access via ES6 Proxy

nojvek:
This is alternate proposal to #62: It should preserve the existing public api as is. Thanks to Proxies we can validate array access.

This is really cool though. Already helped me catch two bugs in our component library. Exactly what I was expecting.

image

### DIFF
--- a/devtools/devtools.js
+++ b/devtools/devtools.js
@@ -41,8 +41,8 @@ function getPanelElementState() {
         debugInfo.appState = panelElem.appState;
       }
 
-      if (panelElem.attrs && Object.keys(panelElem.attrs).length) {
-        debugInfo.attrs = panelElem.attrs;
+      if (panelElem._attrs && Object.keys(panelElem._attrs).length) {
+        debugInfo.attrs = panelElem._attrs;
         debugInfo.attrsSchema = panelElem.constructor.attrsSchema;
       }
 
--- a/lib/component.js
+++ b/lib/component.js
@@ -202,8 +202,21 @@ class Component extends WebComponent {
 
     this.panelID = cuid();
 
-    this.attrs = {};
+    this._attrs = {};
     this._syncAttrs(); // constructor sync ensures default properties are present on this.attrs
+    this.attrs = new Proxy(this._attrs, {
+      get(attrs, attr) {
+        if (attr in attrs) {
+          return attrs[attr];
+        } else {
+          throw new TypeError(`'${attr}' not defined in attrSchema`);
+        }
+      },
+      set() {
+        throw new TypeError(`attrs is readonly`);
+      },
+    });
+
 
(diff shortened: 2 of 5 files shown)


tdumitrescu:
I prefer https://github.com/mixpanel/panel/pull/62 myself. It's more explicit and obvious rather than magical (think python vs ruby), and doesn't have browser compatibility concerns.

nojvek:
#62 
Pros:
 * explicit attr function

Cons:
 * Need a consumer refactor for all this.attrs[attr] -> this.attr(attr) & $attr[attr] -> $attr(attr). Will also need to educate developers about it.
 * Doesn't prevent bad attr access to this.attrs[attr]. Someone may still use the old way and we'll be inconsistent.
 * If we hide this.attrs as this._attrs, do we create new public api to fetch Object.keys/values/entries to do reflection ?

#63
Pros: 
 *  It proxies the actual object so operations such as Object.keys / values/ entries all work as is. Introspection 'attr in this.attrs' works as is too. 
 * No change to public api, developers don't need to learn a new way to do things. They code as they've always done but get stricter validation.

Cons:
 * Browser compatibility. It's an ES6 feature and Proxy is available is last two versions of major browsers. https://caniuse.com/#search=proxy. It's just that Panel tests in really old browsers like Chrome 45. (was released in 2015 - 4 years ago)

tdumitrescu:
> If we hide this.attrs as this._attrs, do we create new public api to fetch Object.keys/values/entries to do reflection ?

Sure. `component.attr()` gets exposed in templates as `$attr()`, and for the exceptional cases where you need the whole map, `component.attrs()` just returns `this._attrs`.

nojvek:
That's a fair alternative too. Gonna wait or a day for couple of people to weigh in. I don't have strong attachments to any proposals.

jbwyme:
:thumbsup: I am pretty indifferent about either approach. I agree with the assessment Noj has laid out but I'd add to the cons of using Proxy:

- It's a big "magical"

Overall, I really don't care which way we go. I'll defer to Ted on the final decision.

andrewpurcell:
I'm pro option 1, `$attr('something')` due to it being less magic and a relatively easy refactor. I think we can easily write an eslint rule to catch `$attrs[]` access.

evnp:
👍 https://github.com/mixpanel/panel/pull/62. Agree this is extremely cool - but in a foundational lib like panel simplicity/clarity/compat seem even more important, worth a refactor. I really don't mind the syntax in https://github.com/mixpanel/panel/pull/62 either, and if attrs are stored under `this._attrs` (instead of `this.attrs`) it's pretty clear this property is not to be accessed directly

nojvek:
Closing this in favor of #62

PR #65, opened 2019-05-02
add pushstate/replacestate events to notify other components of history state changes on navigate

evnp:


lib/router.js
```diff
@@ -49,6 +49,8 @@ export default class Router {
     this.window.history[this.historyMethod] = (...args) => {
```
nojvek: we should probably prevent double proxies. i.e if something calls registerListeners twice it shouldn't create two layers of function calls.

Also would be nice that if the element is removed, it leaves window.history to the original state that it was in.

evnp: agreed, I'll look into this

evnp:
@nojvek took another stab at this, can you PTAL? With the current impl duplicate events will be avoided and history methods will be correctly restored on component disconnect.

lib/router.js
evnp: The current router implementation relies on stacking wrapped history methods, one for each component so that each component's `this.navigate` can be called. I don't see any issue with this, and I'd rather not mess with something so fundamental.

evnp:
note: IE does not support `function.name` so this will not work there; I notice it also doesn't support `new CustomEvent` which we use elsewhere so we may not care. `function.toString` could be used instead if we want this to work in IE.

lib/router.js
```diff
@@ -46,10 +46,22 @@ export default class Router {
...
     this.origChangeStateMethod = this.window.history[this.historyMethod];
-    this.window.history[this.historyMethod] = (...args) => {
-      this.origChangeStateMethod.apply(this.window.history, args);
-      this.navigateToHash();
-    };
+
+    // If this is the true *original* native method, fire pushstate/replacestate events so
+    // that other components can be notified of history changes. Subsequent panel routers
+    // will re-wrap the method but not fire duplicate pushstate/replacestate events.
+    if (this.origChangeStateMethod.name === this.historyMethod) {
```
nojvek: I'm trying to grok how `this.origChangeStateMethod.name === this.historyMethod)` prevents duplicates. is that because the wrapping method has no name, but the original method has a name?

evnp: exactly. open to any alternate means of detecting this, this is the cleanest I could figure out

evnp: to be clear, the original has `name === "pushState"` or `name === "replaceState"`; the wrapped functions don't have this because they are declared as anonymous `(...args) => ...`

nojvek: The way we do it in fetch-cache is something like this.

```
  if (window.__origFetch) {
    // already installed
    return;
  }

  window.__origFetch = window.fetch;
window.fetch = async function fetch(url, options) {
```

We store the origin version as a `__method` name and compare the function. Comparing on name is fine, but feels a bit brittle. I'm okay with it for time being tho. We at-least should write some comments about line 53 for future readers.

@tdumitrescu may have opinions.

evnp: Could we do that here without polluting `window`? Each instance of router only knows of its own `this.origStateChangeMethod` which could be either the native or wrapped version going in; there's no way to compare that to something afaict (unless you store something on window/global, feels more dirty to me). Not sure why you feel this is brittle? [`Function.name`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) is standard across browsers (sans IE as noted earlier)

I can def elaborate on the comments, I thought the 3 lines there would be sufficient to explain the "why" with the code explaining the "how"

evnp: @nojvek added 1e3b861583f4da97f0a105c1b4c2a91b13e18f05 to elucidate

nojvek: Thanks

lib/router.js
```diff
@@ -46,10 +46,26 @@ export default class Router {
...
+    // will re-wrap the method but not fire duplicate pushstate/replacestate events.
+    // Note: this.origChangeStateMethod.name === this.historyMethod is an effective means
+    // of verifying whether this.window.history[this.historyMethod] has been wrapped yet or
+    // not because the original native function will have name === "pushState" / "replaceState",
+    // but the wrapped version will have name === undefined since it is anonymous.
+    if (this.origChangeStateMethod.name === this.historyMethod) {
+      this.window.history[this.historyMethod] = (...args) => {
+        this.origChangeStateMethod.apply(this.window.history, args);
+        this.navigateToHash();
+        window.dispatchEvent(new CustomEvent(this.historyMethod.toLowerCase()));
```
nojvek: smol nit. I'd prefer we not lowercase the event name. pushState and replaceState feel more readable. We already do camelCase for custom events anyway.

evnp: fair, the lowercase was for consistency with native ["popstate"](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event), makes me a little sad that it's lowercase but wdyt?

nojvek: I guess that makes sense then. Let's leave as is. Perhaps a comment?

tdumitrescu:
Hang on can someone give me some context around this? Why do we want to start firing global events rather than scoping to the component in question? Seems like bad behavior.

evnp:
Sure, def wouldn't say this is strictly necessary since we're already making it work within lexicon, the idea is just that it may be useful to other apps instead of using custom app code case-by-case. Use case is this -
- lexicon has a router initialized so it wraps `history.pushState` to run it's `index` route and re-render whenever the url changes
- mp-chrome does _not_ have any router (and I'm not sure how two components with routing on the same page would behave, but even if it works this is a bigger refactor of chrome), so it has a window "popstate" event listener so it can update when the user hits browser back button
- however, there are no native "pushstate"/"replacestate" events, so when we call `app.navigate` from lexicon to switch tabs (there's a button in-app that is supposed to do this, basically a link to another part of the app) mp-chrome has no way of being notified of this url change and cannot re-render appropriately
- the intention of this PR is to add "pushstate"/"replacestate" events on `app.navigate` so mp-chrome can be notified and handle this use case. An earlier version used event bubbling and dispatched the event from the lexicon app element instead of window, I prefer this but am fine with either

tdumitrescu:
Double-routing scenario sounds a little awful to me, I'll trust that you actually need it. But I do think you should be able to fire your custom history event from the component rather than `window`. If you have a real need for it to be a global event, it's real easy to make that happen when initializing an app e.g.
```js
const myApp = document.createElement(`my-app`);
myApp.addEventListener(`pushState`, triggerWindowEventOrDoSomethingGlobal);
```
rather than having Panel add window events.

lib/router.js
tdumitrescu: in other words:
```suggestion
        this.app.dispatchEvent(new CustomEvent(this.historyMethod.toLowerCase()));
```

evnp: I'm definitely on board with this, but how do you feel about adding `{bubbles: true}` to the event? Then (in the concrete example) mp-chrome can add a listener to document instead of getting a reference to the Lexicon element to add the listener. If mp-chrome relies on access to lexicon the order they're initialized becomes important which doesn't seem great - I believe lexicon is initialized 2nd and it seems like we don't want order to matter.

I think the greater issue here is communication between components that don't have an ancestor->descendant relationship in the dom. In this case they both want to do something when url changes - lexicon navigates to a new route, mp-chrome highlights a new report tab - so this event dispatching seems better than having an ancestor component of both receive an explicit event from lexicon and update mp-chrome accordingly. That does seem like the "right" way, but would be so much work when we want such a simple end result here (update component on url change)

tdumitrescu: would this not work for your case?
```js
const app = document.createElement(`lexicon-app`);
app.addEventListener(`pushstate`, () => window.dispatchEvent(new CustomEvent(`pushstate`)));
```

nojvek: +1 on bubbles. Makes it easier to track all navigation changes in one place.

evnp: that wfm yeah. bubbling would be convenient but I can see why that's cleaner as the lib default.

evnp:
@tdumitrescu added 17ac8ad to address your comments, ptal when you get a chance

PR #67, opened 2019-05-24
Make default state available at construction time and flush child state updates

pablofierro:
- Makes default `state` available at construction time so that it can be read on `attributeChangedCallback`.
- Similarly to how `appState` is flushed at `connectedCallback`, flush shared state updates from child component to the parent's state.

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -271,13 +271,13 @@ class Component extends WebComponent {
       this.$panelParent.$panelChildren.add(this);
 
       // share either appState or all of state
-      Object.assign(this.$panelRoot.appState, this.appState); // flush any queued appState changes
-      this.appState = this.$panelRoot.appState;
-      this.isStateShared = this.$panelRoot.isStateShared;
+      // flush any queued appState changes
+      this.appState = Object.assign(this.$panelRoot.appState, this.appState);
 
       // if child element state is shared, point
       // state to parent's state object and flush any
       // queued state changes to the parent state
+      this.isStateShared = this.$panelRoot.isStateShared;
       if (this.isStateShared) {
         this.state = Object.assign(this.$panelRoot.state, this.state);
       }
--- a/lib/component.js
+++ b/lib/component.js
@@ -274,7 +274,13 @@ class Component extends WebComponent {
       Object.assign(this.$panelRoot.appState, this.appState); // flush any queued appState changes
       this.appState = this.$panelRoot.appState;
       this.isStateShared = this.$panelRoot.isStateShared;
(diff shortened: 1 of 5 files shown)


tdumitrescu:
This is a pretty subtle change. Can you add some new test cases that fail without your change to help specify exactly which behavior you're trying to pin down?

lib/component.js
```diff
@@ -217,7 +217,7 @@ class Component extends WebComponent {
     // initialize shared state store, either in `appState` or default to `state`
     // appState and isStateShared of child components will be overwritten by parent/root
     // when the component is connected to the hierarchy
-    this.state = {};
+    this.state = Object.assign({}, this.getConfig(`defaultState`));
```
mduan: Can you also remove the other usage of `defaultState` in `connectedCallback`: https://github.com/mixpanel/panel/blob/06e2194a99e5b105ca3e5349f054060e990ab319/lib/component.js#L288.

pablofierro: 9e78237

pablofierro:
@tdumitrescu I added a new test case in 9e78237 with a more detailed scenario, the only approach I could think of for checking the behavior before/after the change was to break up initialization code from `connectedCallback` into different methods and then stub the method responsible for making the connection of parent and child state what do you think about that approach ?

test/browser/component.js
```diff
@@ -316,6 +316,20 @@ describe(`Nested Component instance`, function() {
       childEl.update({animal: `capybara`});
       expect(el.state.animal).to.equal(`capybara`);
     });
+
+    it.only(`flushes child state updates to parent`, async function() {
```
tdumitrescu:
```suggestion
    it(`flushes child state updates to parent`, async function() {
```

pablofierro: oh my 🤦‍♂ need to do something about that

test/browser/component.js
```diff
@@ -316,6 +316,20 @@ describe(`Nested Component instance`, function() {
       childEl.update({animal: `capybara`});
       expect(el.state.animal).to.equal(`capybara`);
     });
+
+    it(`flushes child state updates to parent`, async function() {
```
tdumitrescu: :+1: other test cases to add:
- verify that this same behavior occurs with `appState`
- full "integration" style test with component created via `.child()`

pablofierro: Thanks for the feedback! I've added other test cases in 7eccefd

tdumitrescu:
Published in v2.1.0, thanks!

PR #68, opened 2019-06-11
dispatch renderError when error occurs during Component._render

andrewpurcell:
Swallowing these exceptions means they're not caught and logged by Rollbar. This was specifically added in #11, so I'm up for some discussion.

This could also be exposed as a hook that child-classes override.


andrewpurcell:
After some thought, I think there ought to be a distinction between components which throw on render error and those that don't.

* If the root component throws (or one of its linked children), the app is probably not recoverable, and should throw an exception.
* If an isolated component in the hierarchy throws, the app might be degraded, but it shouldn't block all rendering (this was probably the original intent here?). However, it should still log this error somewhere beyond the console.

As for how to solve my problem: I could go into all of my components which should block render on error and override `logError` to report but that feels kind of clunky and special-cased.

tdumitrescu:
Yeah, if we take it as a given that we're not always going to throw on render failure, in order to support "degraded but not dead" UI scenarios, then what we want is a mechanism for components to signal when rendering has failed without halting execution. I'd imagine dispatching an event is the obvious solution.

andrewpurcell:
Options:
* at the Component class definition level
* * add a throw-on-error config, default false
* * * you define this configuration at programming time
* * * downside: you don't have a neat place to catch the error (in the rollbar case it just goes to window)
* * add an event-on-error config, default true
* * * downside: have to add on-error type handlers in all callers
* at the subclass level
* * introduce a new base class in our applications which handles reporting
* * * downsides: layer of indirection to add very little, surface for other items to accumulate, no long just a Component subclass
* * introduce a mixin class to override logError
* * * opt-in from all existing code `class MyComponent extends RollbarOnRender(Component)`
* * * downside: still need to remember to opt in?

I think I like the option of adding a mixin to do the override, but my goal would be for all of our components to at least submit rollbars when they fail.

andrewpurcell:
Went with the emit event approach, LMK what you think.

lib/component.js
```diff
@@ -362,8 +362,20 @@ class Component extends WebComponent {
...
+  emitRenderError(message, error) {
+    this.el.dispatchEvent(new CustomEvent(`rendererror`, {
+      detail: {message, error},
+      bubbles: true, composed: true,
+    }));
+  }
+
+  onRenderError(message, error) {
+    this.logRenderError(message, error);
+    this.emitRenderError(message, error);
```
tdumitrescu: This division into 3 single-use functions seems a bit needlessly fussy. Also if it's going to be specifically about errors during rendering then the `message` param is superfluous. How about
```js
onRenderError(error) {
  const message = `Error while rendering ${this.toString()}`;
  console.error(message, this, `\n`, error);
  this.el.dispatchEvent(new CustomEvent(`rendererror`, {
    detail: {message, error},
    bubbles: true, composed: true,
  }));
}
```

test/browser/component.js
```diff
@@ -513,8 +516,13 @@ describe(`Rendering exception`, function() {
   });
 
   it(`logs an error`, function() {
-    expect(el.logError.getCall(0).args[0]).to.contain(`Error while rendering`);
-    expect(el.logError.getCall(0).args[0]).to.contain(`breakable-app`);
+    expect(el.logRenderError.getCall(0).args[0]).to.contain(`Error while rendering breakable-app`);
```
tdumitrescu: I'd say we can do away with this logging test now that there's a test for emitting an event on error. The useful part of this test wasn't to check that something would get written to the console, but rather that it would successfully catch a rendering error and not throw an exception.

lib/component.js
andrewpurcell: the only reason I split it is because the test framework/headless browser fails the test on `console.error`. the test spy was also a stub function before and prevented the call

test/browser/component.js
```diff
@@ -499,22 +499,30 @@ describe(`Nested Component instance with partially shared state`, function() {
 
 describe(`Rendering exception`, function() {
   let el;
+  let eventSpy;
 
   beforeEach(async function() {
+    WCT._config.trackConsoleError = false;
```
andrewpurcell: this is the ugly workaround for console errors (see: https://github.com/Polymer/tools/issues/3080). if there's a better way to do this, lmk

tdumitrescu: If the only point is to stub/spy on console.error then retaining the existing generic `logError()` method implementation is fine. IIRC your previous version converted it into a render-error-specific logger. So the new code would look something like:
```js
logError() {
  console.error(...arguments);
}

onRenderError(error) {
  const message = `Error while rendering ${this.toString()}`;
  this.logError(message, this, `\n`, error);
  this.el.dispatchEvent(new CustomEvent(`rendererror`, {
    detail: {message, error},
    bubbles: true, composed: true,
  }));
}
```
And then anywhere in tests you want to check for console erroring (or prevent WCT from failing on it) you can just stub that one method.

PR #71, opened 2019-07-27
better disconnectedCallback to help with panel memory leak

nojvek:
image

I made a simple test to test that garbage collection is indeed happening.

```ts
import {h, Component, ConfigOptions} from 'panel';

interface AppState {
  showChild: boolean;
  count: number;
}

customElements.define(`mp-blah`, class Blah extends Component<{}, AppState> {
  get config(): ConfigOptions<{}, AppState> {
    return {
      template: ({$component, $app}) => (
        <div attrs={{}}>
          {$app.showChild && $component.child(`mp-blah-child`)}
        </div>
      ),
      defaultState: {
      },
      appState: {
        showChild: false,
        count: 0,
      },
    };
  }
});

customElements.define(`mp-blah-child`, class BlahChild extends Component<{}> {
  get config(): ConfigOptions<{}, AppState> {
    return {
      useShadowDom: true,
      template: ({$app}) => (
        <div attrs={{}}>
          {new Array(100).fill(0).map(() => (
            <div on={{click: this.helpers.handleClick}}>{`${$app && $app.count} `.repeat(30)}</div>
          ))}
        </div>
      ),
      helpers: {
        handleClick: ev => {
          console.log(`click`, this.el, ev);
        },
      },
    };
  }
});

window[`startProfile`] = function startProfile() {
  const intervalId = setInterval(() => {
    const blahEl: any = document.querySelector(`mp-blah`);
    const count = blahEl.appState.count;
    blahEl.updateApp({showChild: false});

    if (count > 100) {
      clearInterval(intervalId);
    } else {
      requestAnimationFrame(() => blahEl.updateApp({showChild: true, count: count + 1}));
    }

  }, 100);
};

document.querySelector(`.app`).innerHTML = `<mp-blah />`;

```

0 DetachedHTMLElements after. Memory stays the same and all the intermediate elements are garbage collected. It does indeed fix the leak 🎉

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -328,8 +328,6 @@ class Component extends WebComponent {
 
     if (this.domPatcher) {
       this.el.removeChild(this.domPatcher.el);
-      this.domPatcher.vnode = null;
-      this.domPatcher.el = null;
       this.domPatcher = null;
       this._rendered = null;
     }
--- a/lib/component.js
+++ b/lib/component.js
@@ -328,13 +328,10 @@ class Component extends WebComponent {
 
     if (this.domPatcher) {
       this.el.removeChild(this.domPatcher.el);
+      this.domPatcher.vnode = null;
+      this.domPatcher.el = null;
       this.domPatcher = null;
       this._rendered = null;
-    }
-
-    if (this.getConfig(`useShadowDom`)) {
-      this.styleTag = null;
-      this.el = null;
     }
 
     this.initialized = false;
--- a/lib/component.js
+++ b/lib/component.js
@@ -332,6 +332,11 @@ class Component extends WebComponent {
       this._rendered = null;
     }
 
+    if (this.getConfig(`useShadowDom`)) {
+      this.styleTag = null;
+      this.el = null;
+    }
+
     this.initialized = false;
   }
 
--- a/lib/component.js
+++ b/lib/component.js
@@ -327,10 +327,7 @@ class Component extends WebComponent {
     }
 
     if (this.domPatcher) {
-      // ensure domPatcher does not lead to Detached HTMLElement references
       this.el.removeChild(this.domPatcher.el);
-      this.domPatcher.state = null;
-      this.domPatcher.renderFunc = null;
       this.domPatcher = null;
       this._rendered = null;
     }
--- a/lib/component.js
+++ b/lib/component.js
@@ -321,6 +321,7 @@ class Component extends WebComponent {
 
     if (this.domPatcher) {
       this.el.removeChild(this.domPatcher.el);
+      this.domPatcher.disconnect();
     }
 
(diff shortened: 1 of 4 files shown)


lib/component.js
```diff
@@ -314,14 +314,15 @@ class Component extends WebComponent {
...
-    }
+
     if (this.domPatcher) {
       this.el.removeChild(this.domPatcher.el);
     }
-    this.domPatcher = null;
-    this.initialized = false;
+
+    // set all self properties to null and leave no dangling references
+    for (const prop of Object.keys(this)) {
```
nojvek: ;( this is a bit too powerful. Gonna re-work this.

lib/component.js
```diff
@@ -314,14 +314,15 @@ class Component extends WebComponent {
     if (this.router) {
       this.router.unregisterListeners();
     }
-    if (this.$panelParent) {
-      this.$panelParent.$panelChildren.delete(this);
```
tdumitrescu: i think removing this line will lead to bad tree behavior. if this change isn't causing tests to fail then we probably need some tests verifying that removing a panel component from the DOM removes it from the panel tree too

nojvek:
It reduces the retainers significantly, doesn't fix everything though. For some reason this/system context is still a retainer and I don't know how to get rid of it. Seems something related to this.fn.bind but I'm not entirely sure. 

This is a pretty safe change that doesn't blow up anything. Tested with main app.

test/browser/component.js
```diff
@@ -126,6 +126,25 @@ describe(`Simple Component instance`, function() {
...
 
+  context(`when detached from DOM`, function() {
+    beforeEach(async function() {
+      document.body.appendChild(el);
+      await nextAnimationFrame();
+      document.body.removeChild(el);
+      await nextAnimationFrame();
+    });
+
+    it(`disconnectedCallback cleans up to be GC friendly`, function() {
```
tdumitrescu: spec docstring format = `"it <does something>"`
```suggestion
    it(`cleans up references to be GC friendly`, function() {
```

test/browser/component.js
```diff
@@ -126,6 +126,25 @@ describe(`Simple Component instance`, function() {
...
+      expect(el.$panelRoot).to.equal(null);
+      expect(el.$panelParent).to.equal(null);
+      expect(el.appState).to.equal(null);
+      expect(el.app).to.equal(null);
+      expect(el.domPatcher).to.equal(null);
+      expect(el._rendered).to.equal(null);
+      expect(el.initialized).to.equal(false);
+    });
```
tdumitrescu: Let's get another test case here exercising the path of a component getting removed and reattached (maybe a couple times?) to ensure that the cleanup and reinit cycle is clean

nojvek: done.

PR #72, opened 2019-07-27
update TemplateScope type definition to match implementation

RyanHirsch:
There is a discrepancy between the type definitions and the implementation for `TemplateScope`. This change updates the type declaration to use `$attr`.

The discrepancy can be seen in this change and the binding in https://github.com/mixpanel/panel/blob/master/lib/component.js#L384

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -68,7 +68,7 @@ declare namespace Component {
     $app: AppState;
 
     /** Attributes parsed from component's html attributes using attrsSchema */
-    $attrs: {[attr: string]: any};
+    $attr: {[attr: string]: any};
 
     /** A reference to the component itself */
     $component: WebComponent;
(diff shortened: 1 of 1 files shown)


lib/index.d.ts
```diff
@@ -68,7 +68,7 @@ declare namespace Component {
     $app: AppState;
 
     /** Attributes parsed from component's html attributes using attrsSchema */
-    $attrs: {[attr: string]: any};
+    $attr: {[attr: string]: any};
```
nojvek: This was my oversight. 

$attr is a function so it should be 

```suggestion
   $attr<A extends keyof Attrs>(attr: A): Attrs[A]; 
```

You’ll need to pass Attr from component generic to here and give ConfigOptions an optional Attrs = AnyAttrs as well.

RyanBuccellato:
Deferring this one to @nojvek

nojvek:
Closing in favor of #73 which updates types and uses them from test fixture to ensure compatibility.

PR #75, opened 2019-09-09
Allow a function to be passed to `update`

RyanHirsch:
This allows a function to be passed to `update` instead of only a partial state object. The function passed is expected to return a partial state object and is given the current state as a parameter.

This behavior will allow for easier integration with libraries like `immer` for state updates, and can potentially allow for more reusable state update utilities because the update function is passed the current state and will not require access to `this`.

```js
this.update(state =>({foo: `updated value derived from ${state.bar}`}));
```

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -174,12 +174,16 @@ class Component extends WebComponent {
    * Applies a state update, triggering a re-render check of the component as
    * well as any other components sharing the same state. This is the primary
    * means of updating the DOM in a Panel application.
-   * @param {object} [stateUpdate={}] - keys and values of entries to update in
+   * @param {object|function} [stateUpdate={}] - keys and values of entries to update in
    * the component's state object
    * @example
    * myWidget.update({name: 'Bob'});
    */
   update(stateUpdate={}) {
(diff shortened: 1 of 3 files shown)


lib/component.js
```diff
@@ -174,12 +174,16 @@ class Component extends WebComponent {
...
    * means of updating the DOM in a Panel application.
-   * @param {object} [stateUpdate={}] - keys and values of entries to update in
+   * @param {object|function} [stateUpdate={}] - keys and values of entries to update in
    * the component's state object
    * @example
    * myWidget.update({name: 'Bob'});
    */
   update(stateUpdate={}) {
+    if (typeof stateUpdate === `function`) {
+      const stateUpdateResult = stateUpdate(this.state);
```
nojvek: How about 
```suggestion
      const stateUpdateResult = (typeof stateUpdate === `function`) ? stateUpdate(this.state): stateUpdate;
      return this._updateStore(stateUpdate, {store: `state`, cascade: this.isStateShared})
```

RyanHirsch:
see #76

PR #85, opened 2020-03-12
boolean fixes for snabbdom upgrade

RyanHirsch:


lib/component.js
```diff
@@ -93,7 +93,7 @@ class Component extends WebComponent {
    * @example
    * {template: state => h('.header', this.child('my-child-widget'))}
    */
-  child(tagName, config={}) {
+  child(tagName, config = {}) {
```
tdumitrescu:
```suggestion
  child(tagName, config={}) {
```
This is the style of the whole lib. If you're going to fiddle with this spacing please do it in a single pass everywhere and make it a separate PR, thanks.

RyanHirsch:
> This is a big lift to support a bad usage; it would be much preferable IMO to update the lib to prevent default-true boolean attrs altogether. Just throw an exception if you try it. We'll be bumping major version because of this snabbdom upgrade anyway.

I agree, but my concern with doing that is all of the places that default to true across the MP product. This is already was already a fairly big lift in regards to test code, as well as hidden regressions due to previous "truthy" values. If we outright do not support this, the upgrade in the product is going to be even harder.

I'm okay with throwing some warnings on `default: true` like we do with the deprecation notices to get the initial upgrade done, then we can do another breaking change to outright ban it.

tdumitrescu:
What would the code of this branch look like if we didn't have to support default-true boolean attrs? My strong preference is not to add a bunch of complexity and extra code to this lib as a result of it; I would honestly rather delay the Snabbdom upgrade altogether until that usage is removed.

RyanHirsch:
> What would the code of this branch look like if we didn't have to support default-true boolean attrs? My strong preference is not to add a bunch of complexity and extra code to this lib as a result of it; I would honestly rather delay the Snabbdom upgrade altogether until that usage is removed.

I propose we cut `Panel@2.7.1` with `snabbdom@0.6.9` (latest before the boolean switch) and add boolean `default: true` warnings with this PR.

tdumitrescu:
SGTM

RyanHirsch:
The farthest we can go is `snabbdom@0.6.5` due to boolean changes. `0.6.6` starts to adhere to a list of "known" booleans (https://github.com/snabbdom/snabbdom/blob/v0.6.5/src/modules/attributes.ts#L8) and only treats them as booleans for the rest of `0.6.x`, then with `0.7.0` the handling is another change for us.

package-lock.json
```diff
@@ -3666,7 +3666,7 @@
         },
         "readable-stream": {
           "version": "1.0.34",
-          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-1.0.34.tgz",
+          "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-1.0.34.tgz",
```
tdumitrescu: This is a vestige of an old npm install from a version that's since been fixed. Can you try deleting your `node_modules` and rerunning `npm install`? IIRC that should avoid regressing these to `http`.

lib/component.js
```diff
@@ -517,6 +517,9 @@ class Component extends WebComponent {
       let attrValue = null;
 
       if (!this.hasAttribute(attr)) {
+        if (attrType === `boolean` && attrSchema.default) {
+          console.warn(`${this}: attr '${attr}' is defaulted to true. boolean attributes default true is deprecated and will be removed with the next release`);
```
tdumitrescu:
```suggestion
          console.warn(`${this}: attr '${attr}' is defaulted to true. Default-true boolean attributes are deprecated and will be removed in a future release`);
```

package-lock.json
RyanHirsch: There apparently gremlins in `npm`. It will always try `https` first. To actually fix, I had to force clean, delete the `package-lock.json` and reinstall.

tdumitrescu: but it doesn't look fixed? these are still marked as http. On latest npm?

RyanHirsch: 🙁 yes. Was on node 12 and did `npm i -g npm` then regenerated the lock file.

tdumitrescu:
published in 2.7.1

PR #88, opened 2020-04-08
controller update

RyanHirsch:
Our base component started supporting function updates with #76 but the same functionality was not given to controlled components.

This change enables controlled components to use `_update` in the same way that standard components can use `update`.


RyanHirsch:
This should be revisited if we do any enhancements to controller components. Right now there isn't a whole lot of active dev using it.

PR #91, opened 2020-05-26
export jsx, similarly to h export

nojvek:
snabbdom-jsx-lite uses snabbdom as a peer dependency. Since panel bundles snabbdom's h, it makes sense to export jsx helper too.


nojvek:
tests fail because of sauce infra issue. Travis doesn't have option of re-running. not sure why

nojvek:
Have ted's slack blessings, so gonna merge 

image

PR #92, opened 2020-05-29
support external template functions with implicitly bound this and make template-hot-loader compatible with es6 modules

nojvek:
support templates as external functions with this implicitly bound. Improved test fixture.

Webcomponent upgrade contains proper type fix for WebComponent.

Make hot loader work with templates as external modules.

### DIFF
--- a/hot/template-loader.js
+++ b/hot/template-loader.js
@@ -14,13 +14,25 @@ module.exports.pitch = function (request) {
     return `module.exports = require(${moduleId});`;
   }
 
+  // hot reloading of templates works by exporting a wrapped closure function that references templateFn
+  // when we receive a new module update, we swap templateFn reference to new template function
+  // this means components that use the template don't need to change their function reference
+  // we just need to call their .update()
   return `
-    let template = require(${moduleId});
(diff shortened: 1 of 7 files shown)


hot/template-loader.js
```diff
@@ -14,13 +14,25 @@ module.exports.pitch = function (request) {
...
+  // we just need to call their .update()
   return `
-    let template = require(${moduleId});
+    const getTemplateFn = (mod) => mod.__esModule ? mod.template : mod;
+    const templateModule = require(${moduleId});
+    let templateFn = getTemplateFn(templateModule);
+    const wrappedTemplateFn = function() {return templateFn.apply(this, arguments)};
     module.hot.accept(${moduleId}, () => {
-      template = require(${moduleId});
+      const templateModule = require(${moduleId});
```
RyanHirsch, line 27: I don't really like the `templateModule` shadowing here but it is not a deal-breaker. We could inline the call to avoid it.

```ts
templateFn = getTemplateFn(require(${moduleId}))
```

nojvek: Makes sense. Eslint would have yelled. Will change.

PR #93, opened 2020-06-10
remove ProxyComponent and ControlledComponent

nojvek:
We usually add more things to panel. This PR makes panel simpler by reducing what it does and offers Component as the core primitive. 

ProxyComponent can be solved by using a jade mixin, or a jsx function that switches which tag is rendered in the template.

StateControllers can be used via composition instead of ControlledComponent inheritance.


andrewpurcell:
I don't buy the rationale for why ProxyComponent is not needed, but I do accept that it's not used anymore and unlikely to be used in the future, so 👍

nojvek:
I guess I don’t fully understand ProxyComponent then. My understanding was that it forwards attrs, events and props to another component based on some dynamic switch.

andrewpurcell:
Yeah, thinking a little bit more about it, this breaks our public API for this package. I don't know if these features were ever adopted outside of Mixpanel, but we ought to have some principle for our philosophy around changes in panel. (If this were in our private repo, I'd be an immediate ✅ , but seeing as I'm not actually blocking anything here, time to discuss.)

nojvek:
>Yeah, thinking a little bit more about it, this breaks our public API for this package. I don't know if these features were ever adopted outside of Mixpanel, but we ought to have some principle for our philosophy around changes in panel. (If this were in our private repo, I'd be an immediate ✅ , but seeing as I'm not actually blocking anything here, time to discuss.)

We try to follow semver, since this is a breaking change, it would be a major version bump. E.g when we upgraded snabbdom, it was breaking boolean attribute behavior and that was a major version bump. 

As for Panel being an OS library, it's a boon and a curse that we are the only dependents. https://www.npmjs.com/package/panel?activeTab=dependents

tdumitrescu:
There are some non-npm uses of it shown by GitHub's "Used by" view, but I wouldn't worry about it. Just bump major version for breaking changes.

PR #94, opened 2020-06-12
Panel Contexts for propagating objects to DOM-descendant components

themacguffinman:
Proposal: https://www.notion.so/mpspeed/Draft-proposal-for-sharing-data-behaviors-and-visual-layouts-with-panel-contexts-4080b6911a4e43d98de7131600d7d21f

This is the ~draft~ implementation of my proposed "Contexts" feature that allows `panel` components to inject arbitrary class dependencies into all descendants at runtime.

I changed my implementation strategy from my initial plans (where contexts are assigned downwards into descendants) to an approach where contexts are pulled from DOM parent nodes by traversing upwards through ancestors.

This is arguably a fairly naive strategy that I had hoped to avoid, suggestions for alternative strategies are welcome.

Edit: the API spec has been updated twice from the draft:
- https://www.notion.so/mpspeed/API-update-1-quick-explainer-da4a6b7ad7a44fdfab2e32d7c3d16136
- https://www.notion.so/mpspeed/API-update-2-quick-explainer-df806e5ceed8428eb0574bcd5d54ecd8

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -768,7 +768,7 @@ class Component extends WebComponent {
     }
   }
 
-  _findNearestContextParent() {
+  _findNearestContextAncestor() {
     if (!this.isConnected) {
       throw new Error(`Cannot determine context before component is connected to the DOM`);
     }
@@ -788,23 +788,23 @@ class Component extends WebComponent {
     return null;
   }
 
-  _findAndMergeActiveContexts() {
-    const contextParent = this._findNearestContextParent();
+  _findAndMergeContextsFromAncestors() {
+    const contextAncestor = this._findNearestContextAncestor();
     const defaultContexts = Object.assign({}, this.getConfig(`defaultContexts`));
 
-    if (contextParent) {
-      // parent contexts must override locally defined defaults
-      return Object.assign(defaultContexts, contextParent._getAvailableContexts());
+    if (contextAncestor) {
+      // ancestor contexts must override locally defined defaults
+      return Object.assign(defaultContexts, contextAncestor._getAvailableContexts());
     }
 
     return defaultContexts;
   }
 
   _getAvailableContexts() {
-    if (!this._cachedActiveContexts) {
-      this._cachedActiveContexts = this._findAndMergeActiveContexts();
+    if (!this._cachedContexts) {
+      this._cachedContexts = this._findAndMergeContextsFromAncestors();
     }
-    return this._cachedActiveContexts;
+    return this._cachedContexts;
   }
 
   getContext(contextName) {
--- a/lib/component.js
+++ b/lib/component.js
@@ -387,7 +387,7 @@ class Component extends WebComponent {
 
     for (const contextName of this.getConfig(`contexts`)) {
       const context = this.getContext(contextName);
-      // Context classes can implement an optional `bindToComponent`` callback that executes each time the component is connected to the DOM
+      // Context classes can implement an optional `bindToComponent` callback that executes each time the component is connected to the DOM
       if (context.bindToComponent) {
         context.bindToComponent(this);
       }
--- a/lib/component.js
+++ b/lib/component.js
@@ -279,6 +279,7 @@ class Component extends WebComponent {
       {},
       {
         css: ``,
+        defaultContexts: new Map(),
         helpers: {},
         routes: {},
         template: () => {
@@ -312,6 +313,25 @@ class Component extends WebComponent {
       throw Error(`"useShadowDom" config option must be set in order to use "css" config.`);
     } else {
       this.el = this;
+    }
+
+    const defaultContexts = this.getConfig(`defaultContexts`);
+    if (!(defaultContexts instanceof Map)) {
+      throw new Error(`"defaultContexts" config option must be a Map`);
+    }
+    for (const [baseContextClass, defaultContext] of defaultContexts.entries()) {
+      if (!baseContextClass || !baseContextClass.constructor) {
+        throw new Error(`"defaultContexts" config option Map must contain class definitions as keys`);
+      }
+      if (!defaultContext || !baseContextClass.constructor) {
+        throw new Error(`"defaultContexts" config option Map must contain class instances as values`);
+      }
+      if (defaultContext.constructor.name === baseContextClass.constructor.name) {
+        throw new Error(`"defaultContexts" config option Map key must be a different class from the class of its value. Please define an abstract base class that declares the context's capabilities and use it as the key.`);
+      }
+      if (!(defaultContext instanceof baseContextClass)) {
+        throw new Error(`"defaultContexts" config option Map value must inherit from the class definition used as the key. Please define a class that implements the capabilities declared by the key class.`)
+      }
     }
   }
 
@@ -512,6 +532,7 @@ class Component extends WebComponent {
             $hooks: hookHelpers,
           }),
         );
+        // this._rendered.data.panelContext = this._getActiveContexts();
       } catch (error) {
         this._logError(`Error while rendering`, this, `\n`, error);
         this.dispatchEvent(
@@ -748,6 +769,41 @@ class Component extends WebComponent {
         }
       }
     }
+  }
+
+  /**
+   * @param {Map<{ new(): any}, any>} contexts - contexts provided by parent panel component
+   */
+  _receiveContexts(contexts) {
+    this._parentContexts = contexts;
+  }
+
+  _getActiveContexts() {
+    const parentContexts = this._parentContexts || new Map();
+    const activeContexts = new Map(parentContexts.entries());
+
+    const defaultContexts = this.getConfig(`defaultContexts`);
+    for (const [baseContextClass, defaultContext] of defaultContexts.entries()) {
+      if (!activeContexts.has(baseContextClass)) {
+        activeContexts.set(baseContextClass, defaultContext);
+      }
+    }
+    return activeContexts;
+  }
+
+  getContext(contextBaseClass) {
+    if (!contextBaseClass) {
+      throw new Error(`@contextBaseClass is null or undefined`);
+    }
+    const defaultContexts = this.getConfig(`defaultContexts`);
+    const defaultContext = defaultContexts.get(contextBaseClass);
+    if (!defaultContext) {
+      throw new Error(`A default context must be declared in config.defaultContexts before this context can be used.`);
+    }
+    if (this._parentContexts && this._parentContexts.has(contextBaseClass)) {
+      return this._parentContexts.get(contextBaseClass);
+    }
+    return defaultContext;
   }
 }
 
--- a/lib/component.js
+++ b/lib/component.js
@@ -811,7 +811,9 @@ class Component extends WebComponent {
     }
 
     if (!(contextName in this._getAvailableContexts())) {
-      throw new Error(`A "${contextName}" context is not available. Check that this component or a DOM ancestor has provided a context in its "defaultContexts" Panel config.`);
+      throw new Error(
+        `A "${contextName}" context is not available. Check that this component or a DOM ancestor has provided a context in its "defaultContexts" Panel config.`,
+      );
     }
 
     return this._getAvailableContexts()[contextName];
--- a/lib/component.js
+++ b/lib/component.js
@@ -280,7 +280,7 @@ class Component extends WebComponent {
       {
         css: ``,
         defaultContexts: {},
-        attachedContexts: [],
+        contexts: [],
         helpers: {},
         routes: {},
         template: () => {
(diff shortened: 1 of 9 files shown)


themacguffinman:
Uh, I'll look at the old browser failures (Chrome 45, Firefox 50, Safari 9, Edge) when I wrap up this PR.

lib/component.js
```diff
@@ -255,6 +256,29 @@ class Component extends WebComponent {
     } else {
       this.el = this;
     }
+
+    const defaultContexts = this.getConfig(`defaultContexts`);
```
andrewpurcell, line 260: so all Contexts have to be instantiated in the constructor, before any data is available to initialize them.

on the one hand, this saves you from the problem of "updating" a context and having to propagate that reference downwards.
on the other hand, once it's set, you can't update it or replace it, so maybe for more complicated contexts, the implementation just delegates behavior to some other object (like the component)

themacguffinman: If you had an example in mind, that would help in discussing the issues here.

I have thought about this, context replacements should not be "managed" by the `panel` framework at all, especially if the desired behavior is complex. Even if you ignore the implementation complications, at best it's a marginally useful helper method, while at worst you have to fight the framework or introduce more context framework concepts to control the context lifecycle to behave how you want. Complicated behavior is better off implemented in the specific scopes in which they are needed rather than centralized and fossilized into the framework.

Contexts shouldn't need to be "replaced" in most cases anyway. Contexts are designed to represent an environment, and you shouldn't need to "replace" your environment mid-way through. You also shouldn't need data to initialize a component context, much in the same way you don't need to "initialize" a browser with customer data: we tell the browser how to obtain data instead (by eg. giving it an API client, or a store knows how to query the API).

lib/component.js
```diff
@@ -255,6 +256,29 @@ class Component extends WebComponent {
...
+    if (!(defaultContexts instanceof Map)) {
+      throw new Error(`"defaultContexts" config option must be a Map`);
+    }
+    for (const [baseContextClass, defaultContext] of defaultContexts.entries()) {
+      if (!baseContextClass || !baseContextClass.constructor) {
```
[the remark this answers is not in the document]
themacguffinman: It's a little awkward for sure, but it's not clear to me how it sucks to a significant extent. Abstract classes and non-functional interfaces are a common concept in OOP and are supported in Typescript. Unfortunately I don't really know of a way to use an interface as a key, but the class hierarchy enforcement is valuable for reasons I note in my proposal: contexts aim to be polymorphic and easily swappable to eliminate the coupling problems of past solutions like globals/singletons/imports/child-components. Symbol-based keys, in contrast, are basically stringly-typed keys that don't establish any relationship between a context key and its implementations.

lib/component.js
```diff
@@ -652,6 +676,68 @@ class Component extends WebComponent {
...
+    if (!contextBaseClass) {
+      throw new Error(`@contextBaseClass is null or undefined`);
+    }
+    const defaultContexts = this.getConfig(`defaultContexts`);
+    const defaultContext = defaultContexts.get(contextBaseClass);
+    if (!defaultContext) {
+      throw new Error(`A default context must be declared in config.defaultContexts before this context can be used.`);
+    }
+
+    return this._getActiveContexts().get(contextBaseClass);
```
nojvek, line 739: Cool. So we only start looking up the ancestor tree only when something calls `getContext`, and then we cache the contexts once they are resolved

lib/component.js
```diff
@@ -255,6 +256,29 @@ class Component extends WebComponent {
     } else {
       this.el = this;
     }
+
+    const defaultContexts = this.getConfig(`defaultContexts`);
+    if (!(defaultContexts instanceof Map)) {
```
nojvek, line 261: I'm not sure what the perf overhead for context related checking is. Since it happens on constructor for every panel component, even though the panel component may have nothing to do with contexts.

nojvek:
My 2 cents. 

Pros:

 - Love that context looks up the ancestor tree rather than depending in $child function like appState. This means things can be treated as good ol custom elements inside each other.
 - Lazy resolving of contexts when getContext is called. 

Cons:
 - I'm not 100% sold on using classes as keys for context map. Seems like a lot of overhead, create base abstract class, create implementation class, every consumer needs base class in default context. Since attrs and state is named, this kind of api feels different to `this.attr('xyz')`. Does something like `this.getContext('logger').error(err)` make more sense? 
 - Would also like this.getContext() to throw an error if the context isn't resolved rather than falling back to a default one. 
 - I like how React has the idea of context providers and consumers. The consumers consume and assume that there will always be a provider somewhere up the tree.
 - We have app/appState and state sharing. They overlap with some of the things that context brings. If we do go ahead with contexts, then I'd love to get to a design where we can kill app/appState and state sharing. Panel becomes a simple library that deals with internal state, state from attrs and an optional context provided somewhere in its dom ancestry.

https://reactjs.org/docs/context.html#examples

The other simple thing about React contexts is that there is only one value. The recommended way is to have one context as an object and consume that. The consumers simply access `this.context`

If we do go via such a route then the component signature becomes simple. `class X extends Component<StateT, AttrsT, ContextT>`

I am not opposed to having multiple named contexts as well, since that's how attrs behaves with `this.attr('xyz')`, having a similar api for contexts i.e `this.context('foo-bar')` seems like less things to learn for a new dev.

lib/component.js
themacguffinman: I can probably replace this particular `instanceof` with `defaultContexts.constructor.name === "Map"`. The other `instanceof` checks only run if the component has contexts. That being said, `instanceof` is a relatively slow operation as I understand it, although I'm not sure how noticeable it is.

nojvek:
@themacguffinman just to be clear, ^ is not a blocking comment. Just my feedback, it's okay not to take it. I am aware that I may have opinions that everyone may not agree with. I am happy to disagree and commit.

Obviously you've put a tremendous amount of effort on the proposal and PR, you're quite enthusiastic to land contexts and do the necessary refactors upstream.

I believe in action, and would be sad to see this stuck in analysis paralysis. 

If @tdumitrescu , @andrewpurcell , @evnp and anyone else who wants to have a stake in contexts are cool with this change, I am cool with it.

themacguffinman:
@nojvek noted. just waiting on some other feedback to see if there's common themes

lib/component.js
```diff
@@ -652,6 +676,68 @@ class Component extends WebComponent {
       }
     }
   }
+
+  _findNearestPanelAncestor() {
```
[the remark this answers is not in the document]
themacguffinman: In this case, "Panel ancestor" means any node in the DOM ancestry that is constructed with the Panel framework. The implementation tries to find Panel components in the DOM ancestry because only Panel components could possibly have Contexts.

Agree that a different name is probably warranted.

lib/component.js
```diff
@@ -652,6 +676,68 @@ class Component extends WebComponent {
...
+  _findAndMergeActiveContexts() {
+    const contextParent = this._findNearestPanelAncestor();
+    const defaultContexts = this.getConfig(`defaultContexts`);
+
+    if (contextParent) {
+      const parentContexts = contextParent._getActiveContexts();
+
+      if (parentContexts.size > 0) {
+        // merge context maps
+        const combinedContexts = new Map(parentContexts.entries());
```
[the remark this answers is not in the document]
themacguffinman: Yes, that's accurate. Contexts must always declare a default, because the component should function even if a production-ready Context is not provided by ancestors.

lib/component.js
```diff
@@ -652,6 +676,68 @@ class Component extends WebComponent {
...
+        for (const [baseContextClass, defaultContext] of defaultContexts.entries()) {
+          if (!combinedContexts.has(baseContextClass)) {
```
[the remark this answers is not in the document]
themacguffinman: No, I there is no way to override parent contexts. The intention is for Contexts to be injected from outside, components should not be depending on specific Context implementations.

lib/component.js
```diff
@@ -255,6 +256,29 @@ class Component extends WebComponent {
...
+    for (const [baseContextClass, defaultContext] of defaultContexts.entries()) {
+      if (!baseContextClass || !baseContextClass.constructor) {
+        throw new Error(`"defaultContexts" config option Map must contain class definitions as keys`);
+      }
+      if (!defaultContext || !baseContextClass.constructor) {
+        throw new Error(`"defaultContexts" config option Map must contain class instances as values`);
+      }
+      if (defaultContext.constructor.name === baseContextClass.constructor.name) {
+        throw new Error(
+          `"defaultContexts" config option Map key must be a different class from the class of its value. Please define an abstract base class that declares the context's capabilities and use it as the key.`,
```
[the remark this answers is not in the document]
themacguffinman: Let's discuss this more informally elsewhere, I think i did a poor job in explaining the "why" behind this, and perhaps there are better ways to achieve my intentions or my intentions may be wrongheaded.

themacguffinman:
Updated with quick explainer: https://www.notion.so/mpspeed/Quick-Panel-Contexts-API-update-explainer-update-1-da4a6b7ad7a44fdfab2e32d7c3d16136

lib/index.d.ts
```diff
@@ -126,7 +138,13 @@ export interface AnyAttrs {
   [attr: string]: any;
 }
 
-export class Component<StateT, AttrsT = AnyAttrs, AppStateT = unknown, AppT = unknown> extends WebComponent {
+export class Component<
+  StateT,
+  AttrsT = AnyAttrs,
+  AppStateT = unknown,
+  AppT = unknown,
+  ContextRegistry = unknown
```
nojvek, line 146:
```suggestion
  ContextRegistryT = unknown
```

test/browser/component.js
```diff
@@ -843,3 +844,212 @@ describe(`Component with required attrs`, function () {
...
+        const widget = document.createElement(`themed-widget`);
+        document.body.appendChild(widget);
+      } catch (err) {
+        // older browsers that rely on a Custom Elements polyfill will throw an error directly
+        errors.push(err.message);
+      }
+      await nextAnimationFrame();
+
+      expect(errors).to.have.lengthOf(1);
+      expect(errors[0]).to.contain(`context is not available`);
```
tdumitrescu, line 993: can we check the more specific error message here? would this be it:
```suggestion
      expect(errors[0]).to.contain(`A "theme" context is not available`);
```

test/fixtures/simple-contexts.d.ts
```diff
@@ -0,0 +1,19 @@
+import {Component, PanelLifecycleContext} from '../../lib';
+
+export interface Theme {
+  getName(): string;
+}
+
+export class LightTheme implements Theme {
+  getName(): 'white';
```
tdumitrescu, line 8: It would be interesting to see a fuller end-to-end test by using these names in component templates and then checking the rendered content. What do you think?

tdumitrescu (approved):
Love it. This is pretty much ready to go. I've offered only a few touchup comments but at this point don't anticipate any substantial changes. Thanks so much for seeing this through!

lib/component.js
```diff
@@ -381,6 +385,13 @@ class Component extends WebComponent {
       this.navigate(window.location.hash);
     }
 
+    for (const contextName of this.getConfig(`contexts`)) {
+      const context = this.getContext(contextName);
+      if (context.bindToComponent) {
```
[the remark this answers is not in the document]
themacguffinman: I chose to run the callbacks on each repeated connect/disconnect because it seems more intuitive and flexible. I added two additional test cases that specifically tests that callbacks are repeatedly executed when the same component is repeatedly connected/disconnected.

test/fixtures/simple-contexts.d.ts
themacguffinman: Makes sense. I rendered the theme name into the class name of widget root divs, then augmented existing tests to check the element class name in addition to the direct `getContext()` checks.

test/browser/component.js
```diff
@@ -843,3 +844,284 @@ describe(`Component with required attrs`, function () {
...
+      document.body.appendChild(app);
+      await nextAnimationFrame();
+      expect(counter.getCount()).to.equal(1);
+      expect(bindToComponentSpy.getCall(0).args[0]).to.equal(app);
+
+      const widget = document.createElement(`slotted-load-counter-widget`);
+      app.appendChild(widget);
+      await nextAnimationFrame();
+      expect(counter.getCount()).to.equal(2);
+      expect(bindToComponentSpy.getCall(1).args[0]).to.equal(widget);
```
tdumitrescu, line 1043: might be useful to precede each of these checks with an explicit verification of the number of calls
```suggestion
      expect(bindToComponentSpy).to.have.callCount(2);
      expect(bindToComponentSpy.getCall(1).args[0]).to.equal(widget);
```
(I think we have sinon-chai here, otherwise the syntax will be a little different)

tdumitrescu (approved):
🎉 🎉 🎉 🎉 🎉  we're good! Thanks for ensuring we have thorough test coverage. It's a simple but powerful feature + implementation, which is just where it should be. Can't wait to start using contexts in app code. :shipit: :shipit: :shipit: :shipit: :shipit:

PR #97, opened 2020-07-21
move AttrsT to 2nd arg and simplify ConfigOptions

nojvek:
The major breaking change is the type signature is now `class Component<StateT, AttrsT = AnyAttrs, AppStateT = unknown, AppT = unknown> extends WebComponent`.

AttrsT is much more frequently used generic type arg, and we want to make things frictionless for simple components declarations.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -49,73 +49,50 @@ export class StateController<State> {
   unsubscribeUpdates(listener: (props: Partial<State>) => void): void;
 }
 
-declare namespace Component {
-  interface Helpers {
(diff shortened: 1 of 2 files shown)


lib/index.d.ts
```diff
@@ -49,73 +49,50 @@ export class StateController<State> {
   unsubscribeUpdates(listener: (props: Partial<State>) => void): void;
 }
 
-declare namespace Component {
```
nojvek, line 52: removing namespace Component because if conflicts with Component (the class) declaration

PR #98, opened 2020-07-25
don't add panelID hash to renderErrors, they break rollbar grouping

nojvek:


tdumitrescu:
I've found it useful when dealing with lists of elements (we didn't add IDs for nothing!). Of course, the most useful output for debugging in most cases is the element reference itself, which chrome devtools helpfully links to the dom inspector. But either way, this is orthogonal to my previous point, which is that rather than modifying the base stringification format for the sake of eventual inclusion in a Rollbar title (weird), we'd be better off replacing the `message` field of the `renderError` event by the bare element reference.

nojvek:
I feel really dumb, I accidentally pushed to master, not realizing I wasn't on the PR branch.

https://github.com/mixpanel/panel/commit/bab38d6d77b7e4eb23ee9731da7050a2b805d5a5

Is there a github setting to disallow pushing to master (and only allow squash merging from PRs) ? It isn't the first time someone's accidentally done this.

nojvek:
Okay reverted master, and moved changed here

tdumitrescu:
Yeah, GH branch protection settings have been getting more fine-grained over the past few years. I've just added a restriction so you can't push directly to master anymore, have to go through a PR with review approval and passing CI.

nojvek:
@tdumitrescu, just an FYI the change you requested for is applied on the PR.

lib/component.js
```diff
@@ -425,11 +425,11 @@ class Component extends WebComponent {
           }),
         );
       } catch (error) {
-        const message = `Error while rendering ${this.toString()}`;
+        const message = `Error while rendering`;
         this._logError(message, this, `\n`, error);
         this.dispatchEvent(
           new CustomEvent(`renderError`, {
-            detail: {message, error},
+            detail: {message, error, component: this},
```
tdumitrescu, line 432:
```suggestion
            detail: {error, component: this},
```
The "message" field now yields absolutely no information not already in the event name "renderError" so let's get rid of it

nojvek:
will publish as minor release

PR #100, opened 2020-07-28
lifecycle helpers and tests

RyanHirsch:
Convenience methods for adding work to be run during the component's `connectedCallback` and `disconnectedCallback` lifecycle methods.

The primary purpose is to make the developer experience better when attaching and tearing down things like event listeners and subscriptions.

`onDisconnected` can be used in a `connectedCallback` to remove attached listeners

Eg:

```ts
connectedCallback() {
if (this.attr(`multi-select`)) {
  // to show shift+click range select feedback
  document.body.addEventListener(`keydown`, this.helpers.shiftKeyListener);
  document.body.addEventListener(`keyup`, this.helpers.shiftKeyListener);
  this.onDisconnected(() => {
    document.body.removeEventListener(`keydown`, this.helpers.shiftKeyListener);
    document.body.removeEventListener(`keyup`, this.helpers.shiftKeyListener);
  });
}
}
```

It also can help ensure that we don't have to store references to the listeners outside `connectedCallback` by leveraging closures

```ts
  connectedCallback() {
    super.connectedCallback();

    const handleResize = () =>
      requestAnimationFrame(() => {
        if (this.tableElement) {
          this.calculateFlowLinks();
        }
      });
    window.addEventListener(`resize`, handleResize);
    this.onDisconnected(() => window.removeEventListener(`resize`, handleResize));

  }
```

`onConnected` is a similar function that can setup and teardown listeners or subscriptions from a constructor. It can also be used by an outside observer like in tests. This use-case is less of an immediate need but felt like the sibling functionality to `onDisconnected`.

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -149,6 +149,31 @@ class Component extends WebComponent {
   }
 
   /**
+   * Enqueue a function to be run at the end of the components connectedCallback. It can optionally return
+   * a function to run during the component's disconnectedCallback
+   */
+  onConnected(fn) {
+    if (this.initialized) {
+      this._maybeEnqueueResult(fn.call(this));
+    }
+    this._connectedQueue.push(fn);
+  }
+
+  _maybeEnqueueResult(result) {
+    if (result && typeof result === `function`) {
+      result.removeAfterExec = true;
+      this._disconnectedQueue.push(result);
+    }
+  }
+
+  /**
+   * Enqueue a function to be run at the end of the components disconnectedCallback.
+   */
+  onDisconnected(fn) {
+    this._disconnectedQueue.push(fn);
+  }
+
+  /**
    * Sets a value in the component's configuration map after element
    * initialization.
    * @param {string} key - key of config item to set
@@ -321,33 +346,14 @@ class Component extends WebComponent {
       updateMode: this.getConfig(`updateSync`) ? `sync` : `async`,
     });
     this.el.appendChild(this.domPatcher.el);
+
+    for (let i = 0; i < this._connectedQueue.length; i++) {
+      const connectedCallbackFn = this._connectedQueue[i];
+      this._maybeEnqueueResult(connectedCallbackFn.call(this));
+    }
+
     this.initialized = true;
     this.initializing = false;
-
-    let connectedCallbackFn = this._connectedQueue.shift();
-    while (connectedCallbackFn) {
-      const result = connectedCallbackFn.call(this);
-      if (result && typeof result === `function`) {
-        this._disconnectedQueue.push(result);
-      }
-
-      connectedCallbackFn = this._connectedQueue.shift();
-    }
-  }
-
-  /**
-   * Enqueue a function to be run at the end of the components connectedCallback. It can optionally return
-   * a function to run during the components disconnectedCallback
-   */
-  onConnected(fn) {
-    if (this.initialized) {
-      const result = fn.call(this);
-      if (result && typeof result === `function`) {
-        this._disconnectedQueue.push(result);
-      }
-    } else {
-      this._connectedQueue.push(fn);
-    }
   }
 
   disconnectedCallback() {
@@ -355,11 +361,12 @@ class Component extends WebComponent {
       return;
     }
 
-    let disconnectedCallbackFn = this._disconnectedQueue.shift();
-    while (disconnectedCallbackFn) {
+    for (let i = 0; i < this._disconnectedQueue.length; i++) {
+      const disconnectedCallbackFn = this._disconnectedQueue[i];
       disconnectedCallbackFn.call(this);
-      disconnectedCallbackFn = this._disconnectedQueue.shift();
     }
+
+    this._disconnectedQueue = this._disconnectedQueue.filter((fn) => !fn.removeAfterExec);
 
     if (this.router) {
       this.router.unregisterListeners();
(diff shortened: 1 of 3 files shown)


lib/component.js
```diff
@@ -361,6 +395,17 @@ class Component extends WebComponent {
     });
   }
 
+  /**
+   * Enqueue a function to be run at the end of the components disconnectedCallback.
+   */
+  onDisconnected(fn) {
```
tdumitrescu, line 401: move `onDisconnected` and `onConnected` alphabetically in the custom public methods section, ~L150 between `navigate` and `setConfig`

lib/index.d.ts
```diff
@@ -222,4 +222,23 @@ export class Component<StateT, AttrsT = AnyAttrs, AppStateT = unknown, AppT = un
    * means of updating the DOM in a Panel application.
    */
   update(stateUpdate?: Partial<StateT> | ((state: StateT) => Partial<StateT>)): void;
+
+  /**
+   * Helper function which will queue a function to be run once the component has been
+   * initialized and added to the DOM. If the component has already had it's connectedCallback
```
tdumitrescu, line 228:
```suggestion
   * initialized and added to the DOM. If the component has already had its connectedCallback
```

lib/index.d.ts
```diff
@@ -222,4 +222,23 @@ export class Component<StateT, AttrsT = AnyAttrs, AppStateT = unknown, AppT = un
...
+   * run, the function will run immediately.
+   *
+   * It can optionally return a function to be queue'd to be run just before the component is
```
tdumitrescu, line 231:
```suggestion
   * It can optionally return a function to be enqueued to be run just before the component is
```

test/browser/component.js
```diff
@@ -686,3 +686,68 @@ context(`$hooks`, function () {
...
+  let el;
+  let spy;
+
+  beforeEach(function () {
+    document.body.innerHTML = ``;
+    el = document.createElement(`simple-app`);
+    spy = sinon.spy();
+  });
+
+  it(`onConnected doesn't fire until connectedCallback`, async function () {
```
tdumitrescu, line 700: you'll want to rejigger these so that "onConnected" and "onDisconnected" are the subjects of `describe` blocks encompassing all of these "it('onConnected..." formulations
```suggestion
  describe(`onConnected`, function() {
    it(`doesn't fire until connectedCallback`, async function () {
```

test/browser/component.js
```diff
@@ -686,3 +686,68 @@ context(`$hooks`, function () {
...
+    document.body.appendChild(el);
+    await nextAnimationFrame();
+
+    el.onConnected(function () {
+      this.update({baz: 42});
+    });
+    await nextAnimationFrame();
+
+    expect(el.textContent).to.contain(`baz: 42`);
+  });
```
tdumitrescu, line 752: to my point above about components getting moved around in the DOM, we'll want test cases in here where the component gets taken out of the DOM and inserted somewhere else (or the same place, doesn't matter, it'll still cause the Custom Elements callbacks to fire).

Once those are working we might want a couple isorender test cases

lib/component.js
```diff
@@ -327,6 +361,13 @@ class Component extends WebComponent {
       return;
     }
 
+    for (let i = 0; i < this._disconnectedQueue.length; i++) {
+      const disconnectedCallbackFn = this._disconnectedQueue[i];
+      disconnectedCallbackFn.call(this);
```
tdumitrescu, line 366: We probably need to try/catch this and L352 above. Since these are arbitrary callbacks we'd like to limit the blast radius if they throw exceptions, and continue to clean up the component gracefully.

lib/component.js
```diff
@@ -148,6 +148,31 @@ class Component extends WebComponent {
     this.$panelRoot.router.navigate(...arguments);
   }
 
+  /**
+   * Enqueue a function to be run at the end of the components connectedCallback. It can optionally return
+   * a function to run during the component's disconnectedCallback
```
tdumitrescu, line 153: jsdoc for the auto-generated API docs is taken from here, not index.d.ts, so at least c/p the full version (with `@param` tag) here

test/browser/component.js
```diff
@@ -686,3 +686,143 @@ context(`$hooks`, function () {
...
+      expect(spy.callCount).to.equal(1);
+
+      document.body.removeChild(el);
+      await nextAnimationFrame();
+      document.body.appendChild(el);
+      await nextAnimationFrame();
+      expect(spy.callCount).to.equal(2);
+    });
+
+    it(`returned function is called each time the component is removed from the DOM`, async function () {
```
tdumitrescu, line 780:
```suggestion
    it(`calls the returned function each time the component is removed from the DOM`, async function () {
```

lib/component.js
```diff
@@ -148,6 +148,39 @@ class Component extends WebComponent {
...
   }
 
+  /**
+   * Helper function which will queue a function to be run once the component has been
+   * initialized and added to the DOM. If the component has already had its connectedCallback
+   * run, the function will run immediately.
+   *
+   * It can optionally return a function to be enqueued to be run just before the component is
+   * removed from the DOM. This occurs during the disconnectedCallback lifecycle.
+   * @param callback Function to be run after the component has been added to the DOM
```
tdumitrescu, line 158: for generated docs, give type and param name:
```suggestion
   * @param {function} fn - callback to be run after the component has been added to the DOM. If this callback returns another function, the returned function will be run when the component disconnects from the DOM.
```
For public APIs it's also nice to give an `@example` usage; see the other methods above and below this.

tdumitrescu:
published in 5.2.0, thanks!

PR #104, opened 2020-08-27
update attrsSchema support required attributes

RyanHirsch:
This change updates the `attrsSchema` to support a new boolean `required` flag. This flag will prevent a default value from being set for the attribute and cause an error to be logged during `connectedCallback`.

Due to the prevention of a default being set, if this attribute is used in a component's template, an exception will be raised and cause a `renderError` to be dispatched as well.

Example usage:
```ts
  static get attrsSchema() {
    return {
      ...super.attrsSchema,
      greeting: {type: `string`, required: true},
    };
  }
```


lib/component.js
```diff
@@ -616,12 +623,19 @@ class Component extends WebComponent {
       const attrType = attrSchema.type;
       let attrValue = null;
 
+      if (attrSchema.deprecatedMsg) {
+        console.warn(`${this}: attr '${attr}' is deprecated. ${attrSchema.deprecatedMsg}`);
+      }
+
```
RyanHirsch, line 629: Moved the deprecation message up due to the new early return for missing required attributes.

lib/component.js
```diff
@@ -327,6 +327,12 @@ class Component extends WebComponent {
     }
     this.initializing = true;
 
+    for (const attrsSchemaKey in this._attrsSchema) {
+      if (!Object.prototype.hasOwnProperty(this._attrs, attrsSchemaKey) && this._attrsSchema[attrsSchemaKey].required) {
+        this._logError(`Missing required attribute ${attrsSchemaKey} for ${this}`);
```
nojvek, line 332: Why not throw real Error ? and go kaboom

If you look at the type check in attrs, that throws a real error, just like other checks in panel.

https://github.com/mixpanel/panel/blob/43b1c4aa1ab48dbcc7d46c0c15ddf6cb4d32f2b1/lib/component.js#L573-L580

lib/component.js
```diff
@@ -616,12 +623,21 @@ class Component extends WebComponent {
       const attrType = attrSchema.type;
       let attrValue = null;
 
+      if (attrSchema.deprecatedMsg) {
+        console.warn(`${this}: attr '${attr}' is deprecated. ${attrSchema.deprecatedMsg}`);
+      }
+
       if (!this.hasAttribute(attr)) {
         if (attrType === `boolean` && attrSchema.default) {
           console.warn(
```
nojvek, line 632: I think this should throw a real error too. We can coalesce into single check


```ts
if (attrType === `boolean` && (attrSchema.default || attrsSchema.required)) {
  throw new Error(`${this}: boolean attr '${attr}' cannot have required or default, since its value is derived from whether dom element has the attribute, not its value`)
}
```

lib/component.js
```diff
@@ -616,12 +623,21 @@ class Component extends WebComponent {
...
             `${this}: attr '${attr}' is defaulted to true. Default-true boolean attributes are deprecated and will be removed in a future release`,
           );
         }
+        if (attrSchema.required && attrType !== `boolean`) {
```
nojvek, line 636: if we throw real error, we don't need this.

lib/index.d.ts
```diff
@@ -116,6 +116,12 @@ export interface AttrSchema {
    * Can be used to auto-generate Attrs interface
    */
   interface?: string;
+
+  /**
+   * Explicitly require an attribute to be passed, useful when no default value can be inferred.
+   * Will log error when missing, and will fail rendering when used in template
```
nojvek, line 122:
```suggestion
```

lib/component.js
RyanHirsch: My thought process was that this approach would allow all missing attributes to be logged, rather than blowing up on the first missing one and having the developer fix them one by one.
In reality, though I doubt this extra logging matters. The scenarios where this has caused me problems have been the addition of single properties. Direct throwing rather than logging and expecting the render to throw an exception is an indirect complexity with no real-world benefits.

lib/index.d.ts
```diff
@@ -116,6 +116,12 @@ export interface AttrSchema {
...
+   */
+  required?: boolean;
 }
```
nojvek, line 125: a pass by suggestion, any chance I can ask to add a 

```
/** 
 For json values, actual tsType e.g `FooBar[]` can be noted here and be used for type & documentation auto generation by other tools.
*/
tsType?: string
```

Your call. I can do separate PR too.

lib/component.js
RyanHirsch: We still need the `if(attrSchema.required)` check, but we can leave off the addition boolean check.

lib/component.js
nojvek: since this is connectedCallback (if the element is part of initial render, the whole dom tree of nested connectedCallbacks will fail with a proper stack trace). We only do try/catch log error in render because it's an asynchronous function call. Even there a render error is both logged and renderError event bubbled up (which could be sent to some error monitoring service e.g rollbar/sentry) 

 _logError doesn't do much. Just logs to console. So there is little way of knowing in the wild if some component is misconfigured.

nojvek:
If you think we're aligned, let's add tests for new logic.

@andrewpurcell you touched some of this code before, you good with the proposed changes?

lib/component.js
```diff
@@ -327,6 +327,12 @@ class Component extends WebComponent {
     }
     this.initializing = true;
 
+    for (const attrsSchemaKey in this._attrsSchema) {
+      if (!Object.prototype.hasOwnProperty(this._attrs, attrsSchemaKey) && this._attrsSchema[attrsSchemaKey].required) {
+        throw new Error(`Missing required attribute '${attrsSchemaKey}' for ${this}`);
```
nojvek, line 332:
```suggestion
        throw new Error(`${this}: Missing required attr '${attrsSchemaKey}'`);
```

mostly so format is consistent with other errors

lib/component.js
```diff
@@ -616,12 +623,21 @@ class Component extends WebComponent {
...
-        if (attrType === `boolean` && attrSchema.default) {
-          console.warn(
-            `${this}: attr '${attr}' is defaulted to true. Default-true boolean attributes are deprecated and will be removed in a future release`,
+        if (attrType === `boolean` && (attrSchema.default || attrsSchema.required)) {
+          throw new Error(
+            `${this}: boolean attr '${attr}' cannot have required or default, since its value is derived from whether dom element has the attribute, not its value`,
           );
         }
+
+        if (attrSchema.required) {
```
nojvek, line 637: could write this as
```
if(!attrSchema.required) {
  attrValue = attrSchema.default;
}
```

test/server/component.js
```diff
@@ -198,4 +200,10 @@ describe(`Server-side component renderer`, function () {
       `Invalid type: bool for attr: bad-attr in attrsSchema. Only ('string' | 'boolean' | 'number' | 'json') is valid.`,
     );
   });
+
+  it(`throws error for missing required attrs`, async function () {
```
nojvek, line 204: Thanks for this.

Let's add a test for boolean default error too.

test/server/component.js
```diff
@@ -198,4 +200,10 @@ describe(`Server-side component renderer`, function () {
...
+    const el = document.createElement(`required-attrs-schema-app`);
+
+    expect(() => el.connectedCallback()).to.throw(`Missing required attribute 'str-attr' for ${el}`);
```
nojvek, line 207:
```suggestion
    expect(() => el.connectedCallback()).to.throw(`${el}: is missing required attr '${attrsSchemaKey}'`);
```

lib/component.js
RyanHirsch: We'd ultimately then assign null to `this._attrs[attr]`

nojvek: talked offline, I think it's ambiguous to have both required and `!this.hasAttribute(attr)`. Rather than return silently, this should throw an Error

RyanHirsch: This is run during the constructor, in which case the attribute may not have been set yet, so we can't error at this point 😞

nojvek: yeah we gotta do the check in connectedCallback or whatever that calls

test/server/component.js
```diff
@@ -7,6 +7,11 @@ import {SimpleApp} from '../fixtures/simple-app';
 import {NestedApp, NestedChild} from '../fixtures/nested-app';
 import {AttrsReflectionApp} from '../fixtures/attrs-reflection-app';
 import {BadAttrsSchemaApp} from '../fixtures/bad-attrs-schema-app';
+import {
+  BadBooleanRequiredAttrsSchemaApp,
+  BadDefaultRequiredAttrsSchemaApp,
+  RequiredAttrsSchemaApp,
+} from '../fixtures/required-attrs-schema-app';
```
nojvek, line 14:
```suggestion
} from '../fixtures/required-attrs-schema-apps';
```

test/server/component.js
```diff
@@ -198,4 +206,31 @@ describe(`Server-side component renderer`, function () {
...
+    const el = document.createElement(`required-attrs-schema-app`);
+
+    expect(() => el.connectedCallback()).to.throw(`${el}: is missing required attr 'str-attr'`);
+  });
+
+  it(`throws error for invalid default required attrs`, async function () {
+    expect(() => {
+      document.createElement(`bad-default-required-attrs-schema-app`);
+    }).to.throw(/attr 'greeting-attr' cannot have both required and default/);
+  });
```
nojvek, line 229: Sweet

test/browser/component.js
```diff
@@ -826,3 +826,20 @@ describe(`Lifecycle Helpers`, function () {
     });
   });
 });
+
+describe(`Required Attrs Components`, function () {
```
nojvek, line 830: I'm not too sure what this test tests. I know @tdumitrescu has stronger opinions on test structure. Please wait on his 👍 /  ✅

tdumitrescu: structurally it's fine, though think about the English statement constructed by the nested levels: "Required Attrs Components renders". Maybe a `context()` would be better, to produce something like "Component / with required attrs / renders successfully when the attrs are provided" and "Component / with required attrs / throws an error when the attrs are not provided". No biggie either way, though it would be good to have the non-happy-path tested in this module.

lib/component.js
```diff
@@ -327,6 +327,15 @@ class Component extends WebComponent {
     }
     this.initializing = true;
 
+    for (const attrsSchemaKey in this._attrsSchema) {
```
tdumitrescu, line 330: Usually we'd use Object.keys(); is there a reason to iterate over non-own keys in this case?
```suggestion
    for (const attrsSchemaKey of Object.keys(this._attrsSchema)) {
```

test/browser/component.js
RyanHirsch: I was struggling to make a non-happy path work. I was wrapping the `.appendChild` with an `expect(...).to.throw` but it would never trip. Do you have any suggestions?

tdumitrescu: yeah presumably in the real browser test the error would be getting thrown asynchronously on connect rather than directly in the appendChild stack, right? if you remove the expect.to.throw does it error out and fail the test at least?

RyanHirsch: it always throws and always fails the test due to the async nature. I was unable to successfully catch it to make an assertion that it failed for the reason I expected.

lib/component.js
RyanHirsch: I was looking at some microbenchmarks and it showed `for...in` was faster. I'm more than happy to follow whatever the convention is

tdumitrescu:
Released in v5.5.0, thanks!

PR #109, opened 2021-01-09
Add example for Contexts

themacguffinman:
Add an example implementation of Contexts.

Additional minor change: small tweak to comment phrasing describing the context lifecycle callbacks in `lib/index.d.ts`.

### DIFF
--- a/examples/contexts/README.md
+++ b/examples/contexts/README.md
@@ -1,6 +1,6 @@
-# Using Contexts to share configuration & state
+# Using Contexts to share configuration and state
 
-A Panel component can be configured with Contexts to share configuration & state in its DOM tree instead of manually passing them into each component as props/attrs.
+A Panel component can be configured with Contexts to share configuration and state in its DOM tree instead of manually passing them into each component as props/attrs.
 
 To configure a component to provide contexts to all components in its DOM tree (including itself), set `defaultContexts` in its `config` object:
 ```js
@@ -29,7 +29,7 @@ get config() {
   return {
     contexts: [`counter`, `darkMode`],
     helpers: {
-      getDarkMode: () => this.getContext(`darkMode`),
+      isDarkModeEnabled: () => this.getContext(`darkMode`),
       getCount: () => this.getContext(`counter`).getCount(),
     },
   };
--- a/examples/contexts/README.md
+++ b/examples/contexts/README.md
@@ -23,10 +23,17 @@ get config() {
 }
 ```
 
-Once you've declared the context names, you can retrieve a context object/value by calling `Component.getContext(contextName)` (or `$component.getContext(contextName)` in jade):
-```jade
--- /* index.jade */
-div Current count is #{$component.getContext(`counter`).getCount()})
+Once you've declared the context names, you can retrieve a context object/value by calling `Component.getContext(contextName)`:
+```js
+get config() {
+  return {
+    contexts: [`counter`, `darkMode`],
+    helpers: {
+      getDarkMode: () => this.getContext(`darkMode`),
+      getCount: () => this.getContext(`counter`).getCount(),
+    },
+  };
+}
 ```
 
 Any object or value can be a context, but you can optionally define callback properties that Panel recognizes and invokes during the component lifecycle:
--- a/examples/contexts/README.md
+++ b/examples/contexts/README.md
@@ -14,7 +14,7 @@ get config() {
 }
 ```
 
-Components must declare the names of any contexts it intends to use in its `config` object:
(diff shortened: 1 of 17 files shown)


examples/contexts/README.md
```diff
@@ -0,0 +1,61 @@
...
+  return {
+    contexts: [`counter`, `darkMode`],
+  };
+}
+```
+
+Once you've declared the context names, you can retrieve a context object/value by calling `Component.getContext(contextName)` (or `$component.getContext(contextName)` in jade):
+```jade
+-- /* index.jade */
+div Current count is #{$component.getContext(`counter`).getCount()})
```
tdumitrescu, line 29: Hmm this bird may have flown the coop in jsx land, but at least in Jade templates I'd recommend passing references via `$helpers` rather than referencing `$component` directly in the template. It helps contain and specify the interface of what's exposed to the template. Back in the bad old days people occasionally tried weird abusive things like making component `update()` calls as part of template rendering code, and we'd been planning to remove the `$component` reference altogether to discourage stuff like that. Hasn't been an issue in a long time but it's why that `$helpers` distinction exists as opposed to just messing with the component/this directly in the template.

examples/contexts/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+Any object or value can be a context, but you can optionally define callback properties you can define that Panel recognizes and invokes during the component lifecycle:
```
tdumitrescu, line 32:
```suggestion
Any object or value can be a context, but you can optionally define callback properties that Panel recognizes and invokes during the component lifecycle:
```

examples/contexts/README.md
```diff
@@ -0,0 +1,61 @@
...
+```js
+class Counter {
+  constructor() {
+    this._count = 0;
+    this._connectedComponents = new Set();
+  }
+  increment() {
+    this._count++;
+    for (const component of this._connectedComponents) {
```
tdumitrescu, line 41: Interesting...it seem to me like this is something that practically any user of `bindToComponent` will want to do (keep an internal collection of components that are using it). I wonder if this should become a basic mixin/class offered directly in Panel. Not a blocker for this PR, just a thought for dev ergonomics.

examples/contexts/contexts.d.ts
```diff
@@ -0,0 +1,15 @@
+// import from the same repo. in a different repo you'd use:
+// import { PanelLifecycleContext } from 'panel';
```
tdumitrescu, line 2:
```suggestion
// import { Component, PanelLifecycleContext } from 'panel';
```

lib/index.d.ts
```diff
@@ -65,9 +65,9 @@ export interface PanelHooks<State> {
 
 // this type is not checked in the Component ContextRegistryT, the Component JS manually checks for these properties instead
 export interface PanelLifecycleContext {
-  // optional callback that executes each time the component is connected to the DOM
+  // optional callback that executes each time a component using this context is connected to the DOM
```
tdumitrescu, line 68: 👍

examples/contexts/README.md
themacguffinman: Yeah food for thought, should definitely keep this in mind if we start seeing this boilerplate frequently. Won't be addressed in this PR.

examples/contexts/README.md
themacguffinman: Makes sense, I cleaned up the examples to use `$helpers` instead.

examples/contexts/README.md
```diff
@@ -0,0 +1,68 @@
+# Using Contexts to share configuration & state
+
+A Panel component can be configured with Contexts to share configuration & state in its DOM tree instead of manually passing them into each component as props/attrs.
```
tdumitrescu, line 3:
```suggestion
A Panel component can be configured with Contexts to share configuration and state in its DOM tree instead of manually passing them into each component as props/attrs.
```

examples/contexts/README.md
```diff
@@ -0,0 +1,68 @@
...
+}
+```
+
+Once you've declared the context names, you can retrieve a context object/value by calling `Component.getContext(contextName)`:
+```js
+get config() {
+  return {
+    contexts: [`counter`, `darkMode`],
+    helpers: {
+      getDarkMode: () => this.getContext(`darkMode`),
```
tdumitrescu, line 32: this is a bool right?
```suggestion
      isDarkModeEnabled: () => this.getContext(`darkMode`),
```

tdumitrescu (approved):
:+1: LGTM, just a couple nitpicks

PR #110, opened 2021-01-11
Set isConnected in shim so that contexts work in iso tests

jakewski:
A component that requires a context can not be tested with iso because the dom-shim does not set `isConnected`

https://github.com/mixpanel/panel/blob/39fabb6cb94bca3809bac83b921e9ad91961f167/lib%2Fcomponent.js#L774-L776


test/server/dom-shims.js
```diff
@@ -41,3 +42,16 @@ describe(`customElement with shadowDom`, function () {
     expect(elem.el.innerHTML).to.equal(`<style>color: blue;</style><div class="foo"><p>Hello</p></div>`);
   });
 });
+
+describe(`appending customElements to DOM`, function () {
+  it(`calls connectedCallback on customElements when using appendChild`, async function () {
+    const elem = document.createElement(`my-app`);
+    const connectedCallbackStub = sinon.stub(elem, `connectedCallback`);
+
+    document.createElement(`div`).appendChild(elem);
```
jakewski, line 51: it's weird making a `div` for this, but it does correctly test the behavior.

I would have preferred to use `document.appendChild` or `document.body.appendChild`, but to do that correctly I will need to submit a PR to `html-element` to make the shim more correct

lib/isorender/dom-shims.js
```diff
@@ -34,6 +34,7 @@ global.requestAnimationFrame = global.requestAnimationFrame || requestAnimationF
     const child = origFunc.apply(this, arguments);
     requestAnimationFrame(() => {
       if (!child.initialized && child.connectedCallback) {
+        child.isConnected = true;
```
tdumitrescu, line 37: Now might be a good time to also patch `removeChild` etc to call `disconnectedCallback` in this environment...

test/server/dom-shims.js
```diff
@@ -1,7 +1,8 @@
 /* eslint-env mocha */
-
+/* eslint-disable no-unused-expressions */ // complains about .to.be.ok
```
tdumitrescu, line 2: can you set this in an `overrides` section of the eslintrc rather than one-off in this file? make it so anything in the test/ dir (or at least test/server/) automatically uses `mocha` env and turns off `no-unused-expressions`

test/server/dom-shims.js
```diff
@@ -41,3 +42,16 @@ describe(`customElement with shadowDom`, function () {
...
+    const elem = document.createElement(`my-app`);
+    const connectedCallbackStub = sinon.stub(elem, `connectedCallback`);
+
+    document.createElement(`div`).appendChild(elem);
+
+    await nextAnimationFrame();
+    expect(connectedCallbackStub.called).to.be.true;
+    expect(elem.isConnected).to.be.true;
+  });
+});
```
tdumitrescu, line 57: TODO server-side test case that uses contexts??

lib/isorender/dom-shims.js
evnp: +1, if you feel like taking it on I'd love to see all of this functionality consolidated here in `dom-shims.js`: https://github.com/mixpanel/analytics/blob/master/iron/common/test/isotests/shims.ts#L363-L391

evnp (approved):

lib/isorender/dom-shims.js
evnp: doesn't need to block this PR fwiw, I'd be happy to collaborate separately if you prefer

jakewski: I'll take a stab at it 👍

jakewski: went ahead and plugged some of our attribute shims, attachShadow, and remove child here https://github.com/mixpanel/panel/pull/110/commits/f4da06345552d6e3ab823c23d3cba625c97225fc per @evnp's linked shims to analytics code. anything else we should add while at it?

test/server/dom-shims.js
jakewski: I took some liberties in unifying to a single `eslintrc` in the root directory and to replace the one one `test/browser`, I imported sinon/expect instead of declaring it global. is this cool? https://github.com/mixpanel/panel/pull/110/commits/3dad3a7d6d597e87ce951f8ddc6aa83e8e58ea0e

test/server/dom-shims.js
jakewski: will look into porting the existing test cases in `browser/`

lib/isorender/dom-shims.js
evnp: Thanks! that's great to have in here

test/server/dom-shims.js
jakewski: ported two basic test cases from `browser/` in https://github.com/mixpanel/panel/pull/110/commits/8ff7019626b452676af8dc5f97cd6d93f4dc85ae

the more advanced cases will need further shimming (things we already take care of in `analytics`) and I didn't want to scope creep this PR any further. lmk how you feel 🙏

lib/isorender/dom-shims.js
```diff
@@ -34,25 +34,51 @@ global.requestAnimationFrame = global.requestAnimationFrame || requestAnimationF
...
 
-  hasAttribute(name) {
-    return !!this.attributes.find((attr) => attr.name === name);
+  removeAttribute(name) {
+    const oldValue = this.getAttribute(name);
+    super.removeAttribute(...arguments);
+    this.__onAttrChanged(name, oldValue, null);
+  }
+
+  hasAttribute (attr) {
```
tdumitrescu, line 73:
```suggestion
  hasAttribute(attr) {
```

test/server/dom-shims.js
```diff
@@ -41,3 +41,39 @@ describe(`customElement with shadowDom`, function () {
...
+    const disconnectedCallbackSpy = sinon.spy(elem, `disconnectedCallback`);
+
+    expect(elem.isConnected).to.not.exist;
+
+    // fake adding to DOM
+    const parent = document.createElement(`div`);
+    parent.appendChild(elem);
+
+    await nextAnimationFrame();
+    expect(connectedCallbackSpy.called).to.be.true;
```
tdumitrescu, line 58: since you've specified sinon-chai as a top-level dep now, may as well turn it on in this module to get assertions like `expect(connectedCallbackSpy).to.have.been.called;`

tdumitrescu (approved):
Nice, just a few nitpicks left. Thanks for all the improvements

lib/isorender/dom-shims.js
jakewski: moved this onto `Element.prototype` for a more general shim after some yarn-link testing (with space before parens)

lib/isorender/dom-shims.js
```diff
@@ -34,25 +34,52 @@ global.requestAnimationFrame = global.requestAnimationFrame || requestAnimationF
...
+    childNode.isConnected = false;
+    childNode.disconnectedCallback();
+  }
+  return child;
+};
+
+Node.DOCUMENT_FRAGMENT_NODE = 11
+
+// html-element does not provide hasAttribute
+Element.prototype.hasAttribute = function (name) {
```
tdumitrescu, line 59: just be consistent with the rest of the file
```suggestion
Element.prototype.hasAttribute = function(name) {
```

tdumitrescu: (PS how is prettier not taking care of this for us?)

jakewski: sorry, I misunderstood the previous suggestion. happy to fix the eslint/prettier autofix setup here or in a later PR

tdumitrescu: yeah no need to block this one on prettier shenanigans

tdumitrescu (approved):

tdumitrescu:
published in 5.7.0 🎩 🎩

PR #115, opened 2021-02-05
Thoroughly walk DOM subtrees to trigger connected/disconected lifecycle in isorender DOM shims

themacguffinman:
The current dom-shim always connects an element during `appendChild`, which shouldn't be semantically correct. A real browser won't connect an element until the whole element subtree is appended to a parent that is already connected, at which point the browser will start traversing the subtree starting from the subtree's root to invoke `connectedCallback` on each node.

In real browsers, the document body is always connected. In the shims we should do the same: the `<body>` tag is considered "always connected" and elements must be appended to a body tag (there should only be one body tag) to trigger the connected lifecycle.

~It's currently an experimental WIP prototype because it seems to break many iso tests in the private mixpanel repo which I have not yet finished investigating; I'm not sure if that is caused by those iso tests relying on incorrect semantics or if these dom-shim changes are behaving incorrectly.~

Note: this is intended to fix errors that are appearing when integrating components that use Contexts into iso tests.

### DIFF
--- a/lib/isorender/dom-shims.js
+++ b/lib/isorender/dom-shims.js
@@ -27,6 +27,7 @@ import requestAnimationFrame from 'raf';
 // is already there
 global.requestAnimationFrame = global.requestAnimationFrame || requestAnimationFrame;
 
+// run a callback for every node of the DOM (sub)tree rooted at the given root node
 function walkDomTree(root, callback) {
   // basic breadth-first tree traversal (non-recursive)
   const breadthQueue = [root];
--- a/lib/isorender/dom-shims.js
+++ b/lib/isorender/dom-shims.js
@@ -33,9 +33,13 @@ function walkDomTree(root, callback) {
   while (breadthQueue.length > 0) {
     const node = breadthQueue.shift();
     callback(node);
-    const children = node.shadowRoot ? node.shadowRoot.childNodes : node.childNodes;
-    for (const child of Array.from(children || [])) {
+    for (const child of Array.from(node.childNodes || [])) {
       breadthQueue.push(child);
+    }
+    if (node.shadowRoot) {
+      for (const child of Array.from(node.shadowRoot.childNodes || [])) {
+        breadthQueue.push(child);
+      }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


themacguffinman:
Related fixes in private repo to prep for this change: https://github.com/mixpanel/analytics/pull/34311

lib/isorender/dom-shims.js
```diff
@@ -27,28 +27,50 @@ import requestAnimationFrame from 'raf';
...
+    const node = breadthQueue.shift();
+    callback(node);
+    for (const child of Array.from(node.childNodes || [])) {
+      breadthQueue.push(child);
+    }
+    if (node.shadowRoot) {
+      for (const child of Array.from(node.shadowRoot.childNodes || [])) {
+        breadthQueue.push(child);
+      }
+    }
```
themacguffinman, line 43: Hmm it might be better to run the connected callback _after_ adding the children to the breadth queue so that children appended during `connectedCallback()` aren't also traversed. I'll give this a go next week.

themacguffinman: Never mind, that doesn't work. Instead, I've just added an if-check to guard against redundant lifecycle calls.

test/server/component.js
```diff
@@ -68,8 +68,9 @@ describe(`Server-side component renderer`, function () {
   });
 
   it(`renders nested components`, async function () {
+    document.body = document.createElement(`body`);
```
tdumitrescu, line 71: Need to do this a lot eh? Maybe it should happen in a `beforeEach`, let tests count on `document.body` being available (and fresh)

lib/isorender/dom-shims.js
```diff
@@ -27,28 +27,54 @@ import requestAnimationFrame from 'raf';
 // is already there
 global.requestAnimationFrame = global.requestAnimationFrame || requestAnimationFrame;
 
+function walkDomTree(root, callback) {
```
tdumitrescu, line 30: let's get an explanatory comment in here e.g.
```suggestion
// run a callback for every node of the DOM (sub)tree rooted
// at the given root node
function walkDomTree(root, callback) {
```

tdumitrescu (approved):
🎉 thanks for fixes

tdumitrescu:
published in v5.8.0, thanks!

PR #125, opened 2021-05-18
Fixes for special cases around root node in template

tdumitrescu:
- If Snabbdom creates a new top-level element rather than patching DOMPatcher's existing element (for instance, when a `key` or tag change forces the creation of a new element), update DOMPatcher's internal element reference to point to it
- If Snabbdom recycles the existing top-level element that DOMPatcher newly created, it won't fire the `insert` hook since it didn't insert any new element into the DOM, so we run the hook from DOMPatcher directly if it's been set on the root node

### DIFF
--- a/lib/dom-patcher.js
+++ b/lib/dom-patcher.js
@@ -56,6 +56,7 @@ export class DOMPatcher {
         insertHook(this.vnode);
       }
     }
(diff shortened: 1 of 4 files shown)


pablofierro (approved):
🚀

PR #128, opened 2021-08-05
bump snabbdom to v3

RyanHirsch:
Required `snabbdom-jsx-lite` bump and dom patcher changes.

Default virtual jade loader settings were also not compatible with this latest version, but the escape hatches were there to make everything work.


lib/dom-patcher.js
```diff
@@ -41,6 +45,9 @@ export class DOMPatcher {
     this.el = document.createElement(tagName);
     if (classMatches) {
       this.el.className = classMatches.map((c) => c.slice(1)).join(` `);
+      // this attribute setting ensures that svg elements behave as expected and will ensure
+      // compatibility with different snabbdom versions
+      this.el.setAttribute(`class`, this.el.classList);
```
RyanHirsch, line 50: this was prompted by `emptyNodeAt` changes.

Original: https://github.com/snabbdom/snabbdom/blob/v0.7.4/src/snabbdom.ts#L63
New: https://github.com/snabbdom/snabbdom/blob/v3.0.3/src/init.ts#L87

RyanHirsch:
CI seems to be bombing out on bad syntax from the transpiled code. It is upset that there is no `=` in the following.

```js
for (const name2 in style.delayed) {
```

tdumitrescu:
It looks like the CI failures are just some pretty old browser versions not supporting some of the transpiled syntax - we should feel free to bump the minimum browser versions in https://github.com/mixpanel/panel/blob/1cac88eae9a205ec493e3eba32cf72432dfc6d7e/wct.conf.json as no one's really running Chrome 45 etc anymore.

RyanHirsch:
@tdumitrescu I've bumped the browser versions and we seem to be green.

examples/animations/webpack.config.js
```diff
@@ -38,7 +38,9 @@ const webpackConfig = {
     root: path.join(__dirname, `node_modules`),
   },
   virtualJadeLoader: {
-    vdom: `snabbdom`,
```
tdumitrescu, line 41: i guess this will work, but would be nicer to update in virtual-jade[-loader]

tdumitrescu (approved):

PR #132, opened 2021-09-13
performance timings

RyanHirsch:
Adds timestamp tracking for component initialization in `connectedCallback` and starts emitting custom events for "slow" renders and component initialization.


lib/index.d.ts
```diff
@@ -195,7 +203,7 @@ export class Component<
    * For use inside view templates, to create a child Panel component nested under this
    * component, which will share its state object and update cycle.
    */
-  child(tagName: string, config?: object): VNode;
+  child<T = object>(tagName: string, config?: T): VNode;
```
RyanHirsch, line 206: Drive-by type update to allow people to opt-in to better typing when using `this.child`

lib/component.js
```diff
@@ -317,6 +317,8 @@ class Component extends WebComponent {
     } else {
       this.el = this;
     }
+
+    this.slowThreshold = 20;
```
tdumitrescu, line 321:
```suggestion
    this.slowThresholdMs = 20;
```
is the idea of making this an instance variable that someone could configure it per component by constructing a component and then modifying `myComponent.slowThreshold` before connecting it to the DOM? maybe it would be better as a `config` value with a default?

lib/dom-patcher.js
```diff
@@ -93,6 +94,16 @@ export class DOMPatcher {
     patch(this.vnode, newVnode);
     this.vnode = newVnode;
     this.el = this.vnode.elm;
+    const elapsedMs = Date.now() - startedAt;
+    if (elapsedMs > this.slowThreshold) {
```
tdumitrescu, line 98: This code is in DOMPatcher, but `slowThreshold` is on Component, isn't it?

RyanHirsch: yeah, good call 🤦

lib/dom-patcher.js
```diff
@@ -33,6 +33,7 @@ export class DOMPatcher {
     this.state = Object.assign({}, initialState);
     this.renderFunc = renderFunc;
     this.vnode = this.renderFunc(this.state);
+    this.component = options.component;
```
RyanHirsch, line 36: alternatively, could pass through the threshold and name directly and dispatch the event on the window.

jordanmnunez (commented):
I love the intention of the changes. However, I am not sure this is the proper API.  I think it is too involved to have config-dependent events. Additionally, what if a component triggers on every renders. That is a problematic amount of events.

I think panel should calculate initializing and render times and then make them accessible via some kind of hook or callback.

lib/component-utils/perf.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Attempt to use a high resolution timestamp when in the browswer environment, but fallback to Date.now
+ * When the performance API is not available.
+ */
+export function getNow() {
```
jordanmnunez, line 5: love this, and i am stealing it for `perf-measure`

PR #138, opened 2022-01-27
Cache & reuse component stylesheets  with Constructible StyleSheets (if supported)

themacguffinman:
Shadow DOM components currently parse & load CSS on every instantiation. Since this CSS is expected to be static across all instances of the same component, we can save work by loading the component stylesheet once and reusing it in subsequent component instances.

Stylesheets are reused using [Constructible StyleSheets](https://developers.google.com/web/updates/2019/02/constructable-stylesheets), which is the best way to reuse stylesheets. One drawback is that it's a very recent proposal that is only implemented in Chrome for now (although [Firefox intends to implement](https://bugzilla.mozilla.org/show_bug.cgi?id=1520690)), but the speed up can be significant in component-heavy pages. The alternative cross-browser approach of using `<link>` with reusable object URLs runs into Flash of Unstyled Content issues that worsen overall performance and triggers visual effects that cannot be fully mitigated from within Panel.

It's hard to pin down a precise impact number, but I estimate that in an [internal component-heavy page](https://mixpanel.com/project/3/view/75/app/insights#report/27717784/new-users-past-7-d) the stylesheet cache is hit 70+ times which consistently saves 50 - 100ms on my x86 development laptop from the overall measured JS + rendering time before chart results are rendered.


test/browser/component.js
```diff
@@ -297,24 +312,24 @@ describe(`Simple Component instance`, function () {
...
 
     it(`applies the styles`, function () {
-      expect(el.shadowRoot.children[0].innerHTML).to.equal(`color: blue;`);
+      expect(getComponentStylesheetText(el)).to.equal(`:host { color: blue; }`);
     });
 
     context(`when applying override styles`, function () {
       it(`appends the overriding styles to the default styles`, async function () {
-        el.setAttribute(`style-override`, `background: red;`);
+        el.setAttribute(`style-override`, `:host { background: red; }`);
```
tdumitrescu, line 324: Is this just so that tests will work cross-browser? We're not making a backwards-incompatible change to the `style-override` API right?

lib/component-utils/styles.js
```diff
@@ -0,0 +1,32 @@
...
+ * Apply CSS to a ShadowRoot, caching the CSS styles by the ShadowRoot host's tag name where possible.
+ * @example
+ * class MyComponent extends HTMLElement {
+ *    constructor() {
+ *      this.el = this.attachShadow({mode: `open`});
+ *      attachStaticStyles(this.el, `:host > * { background: white }`);
+ *    }
+ *  }
+ */
+export function attachStaticStyles(attachedShadowRoot, styleText) {
```
tdumitrescu, line 13: I'm not sure I see the value of adding this whole module with stateless helper plus types file just for the sake of an extra 10 lines of component-init code. I'd rather make it an instance method or better just inline after L319 of component.js and be able to take advantage of the instance context - for instance, rather than reaching up to detect the host tagname, your cache could use the component's Class as a key and not worry about DOM logic.

tdumitrescu (commented):
Nice, this looks like a great improvement, just a couple things to clear up 🍕 🍕 🍕

test/browser/component.js
themacguffinman: Yes that's correct, we're not making any backwards-incompatible changes but the test code for browsers that support `adoptedStylesheets` will balk at `background-red` because is actually _invalid_ CSS, which causes `adoptedStylesheet` to just ignore the CSS altogether.

lib/component-utils/styles.js
themacguffinman: Yeah fair enough, I initially planned to reuse this for `style-override` but the logic for that ended up diverging, it makes sense to just inline it. I'll look into using the component class as cache key instead of tag name.

tdumitrescu (approved):
🚀 🎉 🚀 🎉 🚀

lib/component.js
```diff
@@ -13,6 +13,7 @@ const ATTR_TYPE_DEFAULTS = {
   number: 0,
   json: null,
 };
+const stylesheetCache = new Map(); // key is the tag name, value is a CSSStyleSheet instance
```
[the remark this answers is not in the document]
themacguffinman: whoops, thanks

tdumitrescu:
🌯🌯🌯  published in v5.12.0

PR #151, opened 2022-03-14
Support Constructible Stylesheets in webpack hot module style loader

themacguffinman:
https://github.com/mixpanel/panel/pull/138 broke the webpack hot module style loader which relied on the existence of a `<style>` tag.

### DIFF
--- a/hot/style-loader.js
+++ b/hot/style-loader.js
@@ -20,21 +20,7 @@ module.exports.pitch = function (request) {
       const updatePanelElems = require('panel/hot/update-panel-elems');
       updatePanelElems('${elemName}', elem => {
         if (elem.getConfig('useShadowDom')) {
-          const newStyleText = newStyle.toString();
-          if (elem.el.adoptedStyleSheets) {
-            if (!elem.configStyleSheet) {
-              elem.configStyleSheet = new CSSStyleSheet();
-              elem.el.adoptedStyleSheets = [elem.configStyleSheet, ...this.el.adoptedStyleSheets.slice(1)];
-            }
-            elem.configStyleSheet.replaceSync(newStyleText);
-            elem.setCachedStyleSheet(elem.configStyleSheet);
-          } else {
-            if (!elem.configStyleTag) {
-              elem.configStyleTag = document.createElement('style');
-              elem.el.prepend(elem.configStyleTag);
-            }
-            elem.configStyleTag.textContent = newStyleText;
-          }
+          elem.applyStaticStyle(newStyle.toString(), /* ignoreCache */ true);
           return true;
         }
       });
--- a/hot/style-loader.js
+++ b/hot/style-loader.js
@@ -22,17 +22,18 @@ module.exports.pitch = function (request) {
         if (elem.getConfig('useShadowDom')) {
           const newStyleText = newStyle.toString();
           if (elem.el.adoptedStyleSheets) {
-            if (!elem.getConfig('css')) {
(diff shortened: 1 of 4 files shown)


hot/style-loader.js
```diff
@@ -20,7 +20,18 @@ module.exports.pitch = function (request) {
       const updatePanelElems = require('panel/hot/update-panel-elems');
       updatePanelElems('${elemName}', elem => {
         if (elem.getConfig('useShadowDom')) {
-          elem.el.querySelector('style').textContent = newStyle.toString();
+          const newStyleText = newStyle.toString();
+          if (elem.el.adoptedStyleSheets) {
+            if (newStyleText) {
+              const newStylesheet = new CSSStyleSheet();
+              newStylesheet.replaceSync(newStyleText);
+              elem.el.adoptedStyleSheets = [newStylesheet, ...elem.el.adoptedStyleSheets.slice(1)];
```
hans-lizihan, line 28: why appending rather than changing the stylesheet content here?

themacguffinman: to simplify it so that it implicitly includes the case where the old style was empty (which means there won't be a stylesheet object to modify). There isn't always a stylesheet object there.

themacguffinman: hm the logic still isn't quite right in the case where the old css config is empty but `style-overrides` is not empty....gotta fix that

themacguffinman: just updated the replace sync, sorry for the last minute tweak

tdumitrescu (approved):
It does seem a little complicated/edge-case-laden to have to determine whether or not the element has a custom stylesheet currently, and then moving it on and off the head of the list. I wonder if there's something we could use to mark the custom panel stylesheet, which would make it simple to check which exact state we're in. But at the end of the day, this isn't production code and if it works right now, :shipit:

hot/style-loader.js
hans-lizihan: hmm, one more thing: the cache is not updated. so unmount / mount after hot reload might causing revert of the old style

hans-lizihan (approved):

hot/style-loader.js
themacguffinman: You're right, the `style-override` seemed to be still a bit broken, I've just updated the PR so it should be fixed now. I've also made sure to update the stylesheet cache.

themacguffinman:
@tdumitrescu you're right it's kind of brittle, I've tried to simplify it with straightforward class properties that directly references the static CSS' stylesheet object or style tag so there's no less futzing around with array order.

hot/style-loader.js
```diff
@@ -20,7 +20,21 @@ module.exports.pitch = function (request) {
...
+            }
+            elem.configStyleSheet.replaceSync(newStyleText);
+            elem.setCachedStyleSheet(elem.configStyleSheet);
+          } else {
+            if (!elem.configStyleTag) {
+              elem.configStyleTag = document.createElement('style');
+              elem.el.prepend(elem.configStyleTag);
+            }
+            elem.configStyleTag.textContent = newStyleText;
+          }
```
tdumitrescu, line 37: This is so close to the component code, but not exactly the same. It feels like there must be an opportunity here to make a single publicly-accessible component method that you can use for setting style both in the component code and the hot loader. WDYT?

themacguffinman: Makes sense, I've made an idempotent `Component.applyStaticStyle()` method that should do the right thing.

tdumitrescu (approved):

PR #153, opened 2022-04-22
Updated devtools + use karma to replace wct

hans-lizihan:

PR #157, opened 2022-06-17
Test stability: use specific CSS rule in browser tests to avoid CSS stylesheet expansion

themacguffinman:
This should mitigate potential breakage in new FF versions: https://github.com/mixpanel/panel/pull/153#discussion_r888401300

Replace `background: red;` with `background-color: red;` in tests because some browsers will expand `background: red;` to `background: red none repeat scroll 0% 0%;` in the component stylesheet object, which can cause test failures when we compare input and output style strings.


tdumitrescu (approved):
Thank you!

PR #158, opened 2022-06-21
Improve TypeScript typing for Component.helpers

mduan:
Before this change (in PyCharm) navigating to `Component.helpers.someMethod`'s definition would always take me to https://github.com/mixpanel/panel/blob/74dc4c21b3e1729ae5c38298d0fac64c23e58cc2/lib/index.d.ts#L53. With these typing changes, it'll now navigate to the definition in `Component.config.helpers.someMethod`.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -71,7 +71,7 @@ export interface PanelLifecycleContext {
   unbindFromComponent?(component: Component<any>): void;
 }
 
-export interface ConfigOptions<StateT, AppStateT = unknown, ContextRegistryT = unknown> {
+export interface ConfigOptions<StateT, AppStateT = unknown, ContextRegistryT = unknown, HelpersT extends PanelHelpers = unknown> {
   /** Function transforming state object to virtual dom tree */
   template(scope?: StateT): VNode;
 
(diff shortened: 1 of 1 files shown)


lib/index.d.ts
```diff
@@ -153,7 +153,8 @@ export class Component<
   AttrsT = AnyAttrs,
   AppStateT = unknown,
   AppT = unknown,
-  ContextRegistryT = unknown
+  ContextRegistryT = unknown,
+  HelpersT extends PanelHelpers = unknown
```
tdumitrescu, line 157: In order to take advantage of this, you'll need to include a component-specific helpers type to fill in the generic for each individual component, right? Given that we're moving away from using `helpers` in general (it was mainly there for Jade/Pug ergonomics, and is mostly just extra noise in already-noisy TSX), I wonder if it's worth adding these helpers type defs everywhere. The better move may be to spend that effort moving things out of helpers into main component methods.

tdumitrescu: OK, if it doesn't need anything except this unused generic, I'm fine with adding it, at least until `helpers` goes away. Just get the build passing and we can merge.

hans-lizihan: can we just type it with `&`?

```
get config(): ReturnType<Component['config']> & Helpers
```

another generic on the class is pretty messy indeed.

PR #161, opened 2022-08-10
panel params

hans-lizihan:


lib/component.js
```diff
@@ -805,6 +817,97 @@ class Component extends WebComponent {
...
+    Object.defineProperties(this.params, paramDescription);
+    return this._paramTypes;
+  }
+
+  enqueueUpdate() {
+    if (this._queuedUpdate) {
+      return;
+    }
+    this._queuedUpdate = true;
+    queueMicrotask(() => {
```
hans-lizihan, line 905: `queueMicrotask` will ensure at the end of the tick it's being executed. So we could batch all update to one single `this.update` call

lib/component.js
```diff
@@ -805,6 +817,97 @@ class Component extends WebComponent {
...
+        );
+      }
+
+      if (paramType.default && paramType.required) {
+        throw new Error(`${this}: param '${param}' cannot have both required and default`);
+      }
+
+      if ([Object, Array, Map, Set].includes(type) && !paramType.shouldUpdate) {
+        throw new Error(
+          `Object like params '${param}' on ${this.constructor.name} must have 'shouldUpdate' function passed in`,
```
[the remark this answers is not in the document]
tdumitrescu: "Simple" values should always default to plain equality checks. string A `===` string B. number X `===` number Y. exceptions to that should be very rare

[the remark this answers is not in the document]
tdumitrescu: can you imagine how much boilerplate there would be if every string param had to specify "oh and here's the `shouldUpdate` function, same as every other one: `(a, b) => a === b`"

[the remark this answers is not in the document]
pablofierro: got it yeah that makes sense 👌

lib/component.js
hans-lizihan: browser support for queueMicrotask could be as old as Chrome71 

https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask

could add tests in the browser test suite to verify compatibility.

and core-js provides polyfill as well https://github.com/zloirock/core-js#queuemicrotask

themacguffinman: It was introduced to Safari in version 12.1.

AFAICT, we don't use `core-js`. If babel can handle it, then there shouldn't be a problem.

I guess just make sure browser tests cover it.

lib/component.js
```diff
@@ -805,6 +817,97 @@ class Component extends WebComponent {
...
+    // vuejs like this.data.$data to reference "real" value of params
+    Object.defineProperty(this.params, `$params`, {
+      enumerable: false,
+      value: {},
+    });
+    const paramDescription = Object.keys(paramTypes)
+      .map((param) => [
+        param,
+        {
+          set(p) {
```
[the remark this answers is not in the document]
hans-lizihan: ok, could move to a instance method then

lib/component-utils/paramsModule.js
```diff
@@ -0,0 +1,22 @@
...
+  let params = vnode.data.params;
+  if (!oldParams && !params) return;
+  if (oldParams === params) return;
+  oldParams = oldParams || {};
+  params = params || {};
+  for (key in params) {
+    cur = params[key];
+    old = oldParams[key];
+    if (old !== cur && elm.params) {
+      elm.params[key] = cur;
```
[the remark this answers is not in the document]
hans-lizihan: now i look at it, seems like the `!==` check here is redundant

lib/component.js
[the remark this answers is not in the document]
themacguffinman: Btw I don't intend to be prescriptive on how you implement this, but I just want to suggest that something like `Object.freeze()` could be useful.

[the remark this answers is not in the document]
hans-lizihan: can do `configurable: false`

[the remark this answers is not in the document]
themacguffinman: Yeah that could work too. I do want to note that Object.freeze() also prevents new props from being assigned.

lib/component-utils/paramsModule.js
```diff
@@ -1,21 +1,21 @@
...
-  for (key in params) {
-    cur = params[key];
-    old = oldParams[key];
-    if (old !== cur && elm.params) {
-      elm.params[key] = cur;
-    } else {
-      elm[key] = cur;
+  if (elm._setParams) {
+    elm._setParams(params);
+  } else {
```
themacguffinman, line 12: I forgot to mention this earlier, I don't think there should be a prop fallback. We should never surprise users. If users wanted to use props, they'd use props, params should never quietly behave like props in certain scenarios. We should straight up throw an error if users attempt to apply params to a non-Panel element.

lib/component.js
```diff
@@ -805,6 +817,113 @@ class Component extends WebComponent {
...
+
+  _enqueueParamUpdate(param, value) {
+    this._count = 0;
+    if (this._queuedParamUpdates) {
+      this._queuedParamUpdates[param] = value;
+      return;
+    }
+    this._queuedParamUpdates = {[param]: value};
+    queueMicrotask(() => {
+      const options = {
```
hans-lizihan, line 900: @themacguffinman updated a bit of update mechanism for here. to make sure `this.params` points to the old value, I need to run hooks and update manually here.

hans-lizihan:
@themacguffinman the type thing are updated as well

lib/component.js
```diff
@@ -650,7 +664,7 @@ class Component extends WebComponent {
 
     const hook = (this.getConfig(`hooks`) || {})[hookName];
     if (hook) {
-      hook(...params);
+      hook.apply(this, params);
```
hans-lizihan, line 667: just make sure the hook has `this` correctly binded

themacguffinman: I'm inclined to avoid changing this as a drive-by change. At this point it's long-standing behavior that someone may rely upon, and it's also not especially necessary. If people want `this` binded, they can easily use arrow functions.

lib/component-utils/paramsModule.js
```diff
@@ -0,0 +1,11 @@
+function updateParams(oldVnode, vnode) {
+  const elm = vnode.elm;
+  const oldParams = oldVnode.data.params;
+  const params = vnode.data.params;
+  if (!oldParams && !params) return;
+  if (oldParams === params) return;
```
themacguffinman, line 6: This module shouldn't have these if-checks. All null-checks and equality checks should be in the `_setParams` method. Any comparisons between the old & new `vnode.data.params` should be irrelevant to Panel params, the Panel Component should have sole responsibility for when and how new params are applied.

lib/component.js
```diff
@@ -805,6 +819,114 @@ class Component extends WebComponent {
...
+
+      if ([Object, Array, Map, Set].includes(type) && !paramType.shouldUpdate) {
+        throw new Error(
+          `Object like params '${param}' on ${this.constructor.name} must have 'shouldUpdate' function passed in`,
+        );
+      }
+
+      const paramSchemaObj = {
+        type,
+        default: paramType.default ? paramType.default : PARAM_TYPE_DEFAULTS.get(type),
```
themacguffinman, line 864: `ParamSchema.default` is allowed to be undefined, it should not fallback to `PARAM_TYPE_DEFAULTS.get(type)`. When the default is undefined, it means that the param is required and will cause errors if the component’s params are updated without a setting a value for this param.

lib/component.js
```diff
@@ -805,6 +819,114 @@ class Component extends WebComponent {
...
+        required: paramType.required ? paramType.required : false,
```
themacguffinman, line 865: `required` should not be a schema property, whether a param is required should be inferred from the `default` property. If `default` is undefined, it means that the param is required and will cause errors if the component’s params are updated without a setting a value for this param.

lib/component.js
```diff
@@ -805,6 +819,114 @@ class Component extends WebComponent {
...
+        // by default use === for comparision
+        shouldUpdate: paramType.shouldUpdate || ((oldVal, newVal) => oldVal !== newVal),
+      };
+
+      // set default value for the params
+      this.params.$params[param] = paramSchemaObj.default;
+
+      this._paramTypes[param] = paramSchemaObj;
```
themacguffinman, line 873: You should use names that distinguish variables & properties that store param schema objects from the variables & properties that store the actual param type.

For example, `this._paramTypes` actually stores param schemas, not types, so you should call it `this._paramSchemas`. Likewise, when you declare a variable named `paramType`, you should not use it to store param schemas `paramType = {type: paramType};`.

lib/component.js
```diff
@@ -805,6 +819,114 @@ class Component extends WebComponent {
...
+          `Object like params '${param}' on ${this.constructor.name} must have 'shouldUpdate' function passed in`,
+        );
+      }
+
+      const paramSchemaObj = {
+        type,
+        default: paramType.default ? paramType.default : PARAM_TYPE_DEFAULTS.get(type),
+        required: paramType.required ? paramType.required : false,
+        // by default use === for comparision
+        shouldUpdate: paramType.shouldUpdate || ((oldVal, newVal) => oldVal !== newVal),
```
themacguffinman, line 867: There's no need to declare & bind a new arrow function `(oldVal, newVal) => oldVal !== newVal` for every param that doesn't define `shouldUpdate`. There should be one non-arrow function `function defaultShouldUpdate(oldVal, newVal) { return oldVal !== newVal; }` declared once which is then reused each time it's needed.

lib/component.js
themacguffinman: It's not clear to me that `this._enqueueParamUpdate()` should exist as a separate method, nor is there a need to use `queueMicrotask()`. Just run the hooks inside `_setParams()`.

lib/component.js
```diff
@@ -805,6 +819,114 @@ class Component extends WebComponent {
...
+      };
+      this.runHook(`preUpdate`, options, this.state, Object.assign({}, this.params.$params, this._queuedParamUpdates));
+      Object.assign(this.params.$params, this._queuedParamUpdates);
+      this.domPatcher.update(this.state);
+      this.runHook(`postUpdate`, options, this.state, this.params);
+      this._queuedParamUpdates = null;
+    });
+  }
+
+  _setParams(params) {
```
themacguffinman, line 914: Is there a particular reason why you've made the `_setParams()` method effectively private but prefixing it with an underscore `_`? I think it's a good idea to make it a public method so that non-JSX code can imperatively use it as well.

lib/index.js
```diff
@@ -15,7 +15,11 @@ import {jsx} from 'snabbdom-jsx-lite';
 
 const {StateController, StateStore} = ComponentUtils;
 
+const Panel = Component;
+
 export {
+  /** {@link Component} newer class with modified generic types */
```
themacguffinman, line 21: Hm IIUC, this should be `@link Panel` not `@link Component`

lib/component.js
hans-lizihan: hmm a quick search in the codebase all hooks are using arrow function. i could revert it back to see if my testcase could pass

themacguffinman: This is technically a public project/package. I don't think we should break long-standing behavior unless necessary. I'd recommend you revert it back.

hans-lizihan: https://github.com/mixpanel/panel/pull/161/files#diff-094bb12e1d2490b9c34b42b64c7e2f9fb0b84fd96d4f42eb2cee4e31296453adR85 the challenge is there's a testcase targeting the hook `this` context, i can't get it work without tweaking this apply thing

themacguffinman: You should be able to reframe that test code without `this` binding:
```typescript
    child._config.hooks = {
      preUpdate(stateUpdate, paramsUpdate) {
        expect(child.params.str).to.equal(``);
        expect(paramsUpdate.str).to.equal(`abc`);
      },
    };
```

lib/component.js
hans-lizihan: yeah now that i think twice, seem it's not possible to have 2 passes of `domPatcher.patch` to update params, so a simple method could do the trick

lib/component.js
hans-lizihan: ah I was trapped in my own little mind maze, thx for the help 🙈

lib/component.js
```diff
@@ -805,6 +819,109 @@ class Component extends WebComponent {
...
+      const newParam = params[param];
+      const oldParam = this.params[param];
+      if (this._paramTypes[param].shouldUpdate(oldParam, newParam)) {
+        if (this._paramTypes[param].required && (newParam === undefined || newParam === null)) {
+          throw new Error(`${param} is required, ${newParam} passed in`);
+        }
+        toUpdate[param] = newParam;
+      }
+    }
+    if (this.initialized) {
```
hans-lizihan, line 913: @themacguffinman this is what i've seen. so if no `enqueueMicroTask`, the `DOMPathcer` sometimes is not ready yet when _setParams is executed.

lib/component.js
hans-lizihan: hmm i assume that except for panel, no one should use this method to set params, even want to use a symbo further hide it. what's the usecase for exposing this method outside?

lib/component.js
hans-lizihan: hmm personally i think implicitly making a param required without default value is kinda tricky. Those logic for required and default basically inherits from the attr schema.

lib/component.js
themacguffinman: Eg you have a test case:
```typescript
it(`renders line series`, function () {
  const myChart = new InsightsLineChart();
  document.body.appendChild(myChart);
  expect(myChart.state.series).to.equal([]);
  myChart.setParams({bookmark: INSIGHTS_BOOKMARK, results: INSIGHTS_QUERY_RESULTS});
  expect(myChart.state.series).to.be.of.length(6);
});
```

Sometimes we may want to manually create & attach elements in the DOM, eg
```typescript
downloadInvoicePDF(url) {
  const link = document.createElement(`mp-download-link`);

  // launch named download via hidden link
  link.setParams({href: url, download: `invoice.pdf`});
  link.click();
}
```

lib/component.js
themacguffinman: That seems like the right behavior. It's what `this.update()` does. If you run `this.update()` before the component is initialized, it just skips the hook.

https://github.com/mixpanel/panel/blob/74dc4c21b3e1729ae5c38298d0fac64c23e58cc2/lib/component.js#L812-L827

lib/component.js
hans-lizihan: hmm ok that makes sense, though the downloadPDF thing seems an anti pattern for me 🙈 

for the testcases, i think it should be sth like this (passing in the para with `init`)

```
it(`renders line series`, function () { 
  const comp = init(`insight-chart`, {
    params: {
      bookmark: INSIGHTS_BOOKMARK,
      results: INSIGHTS_QUERY_RESULTS,
    }
  });
  const myChart = mount(comp)
  expect(myChart.state.series).to.equal([]); 
  expect(myChart.state.series).to.be.of.length(6); 
});
```

themacguffinman: Sure but `init()` itself is going to need to call `component.setParams()` to apply params because params is an artificial Panel concept.

Also, it seems likely that tests will want to change the params during a test case. Your own code example simultaneously expects `myChart.state.series` to be equal to an empty array `[]` and be `of.length(6)`. Those two expect statements were supposed to separately check conditions before & after a params change

hans-lizihan: right just make `_setParams` to `setParams` to indicate it's not strictly private method

themacguffinman: Yeah. Generally nothing outside Panel's own code should be using methods prefixed with underscores because the prefix conventionally means Panel can change or break compatibility with these private methods at any time.

lib/component.js
hans-lizihan: just add explicit check for `paramSchema.default !== undefined` for this. it will fallback to empty value for those primitive types only if no `default` key is provided to the param schema. 

technically when a value is declared then it's value won't be undefined, at least it should be `null`.

themacguffinman: Hm thinking about it a bit more, I think this part of the Panel Params spec is wacky. There shouldn't be a `default` for each individual param schema, defaults should be done the same way React does it and how Panel does default states: a single dictionary containing all defaults.

Default values should be declared all at once in config like this:
```typescript
get config() {
  return {
    ...
    defaultParams: {
      x: 1,
      y: 2,
    },
  };
}
```

Then each param schema should only have a `required` boolean, no `default`.

hans-lizihan: hmm could do it to keep close to react.

themacguffinman: But either way, there should be no defaults for default values. If they don't specify a default for a certain param, the value _should be undefined_. It's not clear that there are always good defaults for parameters, for eg a component may work differently if a function param is not specified, just quietly filling in a no-op function as a supposedly "sensible default" is not desirable behavior.

lib/component.js
```diff
@@ -805,6 +826,89 @@ class Component extends WebComponent {
...
+  get params() {
+    return this._params;
+  }
+
+  setParams(params) {
+    // if nothing change, then we return in advance to avoid unecessary update
+    if (Object.keys(params) === 0) {
+      return;
+    }
+    if (!this.shouldComponentUpdate(params, this.state)) {
```
themacguffinman, line 885: `this.shouldComponentUpdate()` shouldn't prevent params from being assigned, it should only prevent hooks and DOM updates from firing. The argument params are _always_ applied regardless of what `shouldComponentUpdate()` returns.

That's the way React props work, and it makes sense to do it this way because it's likely you still want to compare the next set of params to the previously applied params even if the previously applied params didn't trigger an DOM update, you never want to discard the previously applied params.

lib/component.js
```diff
@@ -805,6 +826,89 @@ class Component extends WebComponent {
...
+    for (const param in params) {
+      if (!this._paramSchemas[param]) {
+        throw new Error(`extra param '${param}' on ${this.constructor.name} is not defined in schema`);
+      }
+      const newParam = params[param];
+      if (this._paramSchemas[param].required && (newParam === undefined || newParam === null)) {
+        throw new Error(`param '${param}' in ${this.constructor.name} is required, ${newParam} passed in`);
+      }
+    }
+    const newParams = Object.assign({}, this.params, params);
```
themacguffinman, line 900: Like I mentioned before, we're not updating the existing params, we're replacing params each time.
```suggestion
    const newParams = Object.assign({}, params);
```

lib/component-utils/paramsModule.js
```diff
@@ -0,0 +1,7 @@
+function updateParams(oldVnode, vnode) {
+  const elm = vnode.elm;
+  if (elm.setParams) {
+    elm.setParams(vnode.data.params || {});
+  }
```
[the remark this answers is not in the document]
themacguffinman: There should be a test case for this

lib/component.js
```diff
@@ -805,6 +826,89 @@ class Component extends WebComponent {
...
+    return this._paramSchemas;
+  }
+
+  get params() {
+    return this._params;
+  }
+
+  setParams(params) {
+    // if nothing change, then we return in advance to avoid unecessary update
+    if (Object.keys(params) === 0) {
```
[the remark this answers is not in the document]
themacguffinman: There should be a test case enforcing this.

lib/component.js
```diff
@@ -805,6 +826,89 @@ class Component extends WebComponent {
...
+      return;
+    }
+    if (!this.shouldComponentUpdate(params, this.state)) {
+      return;
+    }
+    const updateOptions = {
+      cascade: false,
+    };
+    for (const param in params) {
```
tdumitrescu, line 891:
```suggestion
    for (const [paramName, newValue] of Object.entries(params)) {
```

lib/component.js
```diff
@@ -805,6 +826,89 @@ class Component extends WebComponent {
...
+      if (!this._paramSchemas[param]) {
+        throw new Error(`extra param '${param}' on ${this.constructor.name} is not defined in schema`);
+      }
+      const newParam = params[param];
+      if (this._paramSchemas[param].required && (newParam === undefined || newParam === null)) {
+        throw new Error(`param '${param}' in ${this.constructor.name} is required, ${newParam} passed in`);
```
tdumitrescu, line 897: so `required` means "not null/undefined" rather than "must exist in the `params` object"?

lib/component.js
```diff
@@ -852,7 +956,7 @@ class Component extends WebComponent {
 
     const {store, cascade} = options;
     Object.assign(this[store], stateUpdate);
-    if (store !== `state` || this.shouldUpdate(this[store])) {
+    if (store !== `state` || this.shouldComponentUpdate(null, this[store])) {
```
tdumitrescu, line 959: another breaking change?

lib/index.d.ts
```diff
@@ -228,14 +259,52 @@ export class Component<
...
+   *
+   * To be overridden by subclasses, defining conditional logic for whether
+   * a component should rerender its template given the state and params to be applied.
+   * In most cases this method can be left untouched, but can provide improved
+   * performance when dealing with very many DOM elements.
+   *
+   * @example
+   * shouldComponentUpdate(params, state) {
+   *   if (params.bookmark.id !== this.params.bookmark.id) {
+   *     return false;
```
tdumitrescu, line 291: In this example, shouldn't it `return true` in the case where the params update is changing it to a totally different bookmark?

lib/index.d.ts
```diff
@@ -295,3 +356,15 @@ export class Component<
    */
   getContext<ContextKey extends keyof ContextRegistryT>(contextName: ContextKey): ContextRegistryT[ContextKey];
 }
+
+/**
+ * Panel component that only accepts 3 generic types
+ */
+export class Panel<ParamT = unknown, StateT = unknown, ContextRegistryT = unknown> extends Component<
```
tdumitrescu, line 363: `Panel` is a bit of an odd name for it, how about something explicit like `ParamComponent`

package.json
```diff
@@ -44,11 +44,13 @@
   },
   "homepage": "https://github.com/mixpanel/panel",
   "dependencies": {
+    "core-js": "3.24.1",
```
tdumitrescu, line 47: This and `regenerator-runtime` are only needed for running test environment, so move them to `devDependencies`

test/browser/index.js
```diff
@@ -3,12 +3,14 @@ import sinonChai from 'sinon-chai';
 
 chai.use(sinonChai);
 
-import 'babel-polyfill';
+import 'core-js';
+import 'regenerator-runtime';
```
tdumitrescu, line 7: Why do we need this btw? Surely all the browsers we need to support have async/await natively

test/browser/index.js
```diff
@@ -3,12 +3,14 @@ import sinonChai from 'sinon-chai';
...
 
 import '../fixtures'; // import fixtures
 
 // import tests
 import './component';
 import './router';
+import './component-param';
```
tdumitrescu, line 14:
```suggestion
import './param-component';
```
(also fits with my naming suggestion above)

test/iso/component-param.js
```diff
@@ -0,0 +1,179 @@
...
+
+customElements.define(`param-child`, ParamChild);
+customElements.define(`param-parent-app`, ParamParentApp);
+customElements.define(`param-required-and-default-app`, DefaultAndRequiredParam);
+customElements.define(`extra-param-pass-in-child`, ExtraParamPassInChild);
+customElements.define(`non-primitive-type-param-class`, NonPrimitiveTypeParamClass);
+customElements.define(`non-primitive-type-param-string`, NonPrimitiveTypeParamString);
+customElements.define(`should-component-update-params-app`, ShouldComponentUpdateParamsApp);
+
+describe(`panel-params`, () => {
```
tdumitrescu, line 22:
```suggestion
describe(`Panel params`, function() {
```
change arrow functions to normal functions for mocha callbacks throughout

test/iso/component-param.js
```diff
@@ -0,0 +1,179 @@
...
+        <div id="obj">{}</div>
+        <div id="arr">[]</div>
+        <div id="map">{}</div>
+        <div id="set">{}</div>
+      </div>
+    `),
+    );
+  });
+
+  it(`updates parent will cause update in child`, async () => {
```
tdumitrescu, line 48: more grammatical, something like
```suggestion
  it(`updates child whenever parent is updated`, async function() {
```

test/iso/component-param.js
```diff
@@ -0,0 +1,179 @@
...
+  it(`respects defaultParams`, async () => {
+    const el = new ParamParentApp();
+    el.connectedCallback();
+    document.body.appendChild(el);
+    await nextAnimationFrame();
+    expect(el.childNodes[0].params.defaultString).to.equal(`defaultString`);
+    expect(el.childNodes[0].params.noDefaultString).to.equal(undefined);
+  });
+
+  it(`respects required field`, () => {
```
tdumitrescu, line 106:
```suggestion
  it(`validates presence of required field`, function() {
```

test/iso/component-param.js
```diff
@@ -0,0 +1,179 @@
...
+    el.setConfig(`updateSync`, true);
+    el.connectedCallback();
+    document.body.appendChild(el);
+    expect(() => el.update({requiredString: undefined})).to.throw(
+      Error,
+      `param 'requiredString' in ParamChild is required, undefined passed in`,
+    );
+  });
+
+  it(`respects required field`, () => {
```
tdumitrescu, line 117:
```suggestion
  it(`disallows combination of required field and default value`, function() {
```

test/iso/component-param.js
```diff
@@ -0,0 +1,179 @@
...
+  });
+
+  it(`throws error if non primitive used in param schema(string)`, () => {
+    expect(() => new NonPrimitiveTypeParamString()).to.throw(
+      Error,
+      `Invalid type: json for param: A in paramSchema. Only ('Array' | 'String' | 'Boolean' | 'Number' | 'Object' | 'Function' | 'Map' | 'Set') is valid.`,
+    );
+  });
+
+  it(`throws error if unkown param pass in child`, () => {
```
tdumitrescu, line 138:
```suggestion
  it(`throws error if unkown param is passed in child`, function() {
```

test/iso/component-param.js
```diff
@@ -0,0 +1,179 @@
...
+    const el = new ExtraParamPassInChild();
+    el.setConfig(`updateSync`, true);
+    expect(() => el.connectedCallback()).to.throw(Error, `extra param 'extra' on ParamChild is not defined in schema`);
+  });
+
+  it(`hooks will be run`, async () => {
```
tdumitrescu, line 144:
```suggestion
  it(`runs hooks`, async function() {
```

test/iso/component-param.js
```diff
@@ -0,0 +1,179 @@
...
+    await nextAnimationFrame();
+    await nextAnimationFrame();
+    expect(child.getConfig(`hooks`).preUpdate.callCount).to.equal(1);
+    // no state for the child
+    expect(child.getConfig(`hooks`).preUpdate.firstCall.args[0]).to.equal(null);
+    // preUpdate hook
+    expect(child.getConfig(`hooks`).preUpdate.firstCall.args[1].str).to.equal(`abc`);
+    expect(child.getConfig(`hooks`).preUpdate.firstCall.args[1].num).to.equal(5);
+    expect(child.getConfig(`hooks`).preUpdate.firstCall.args[1].bool).to.equal(true);
+    sinon.restore();
```
tdumitrescu, line 161: This won't get called if the test case throws an exception before this line. That's why you generally use `beforeEach`/`afterEach` for this kind of cleanup

test/iso/component-param.js
```diff
@@ -0,0 +1,179 @@
...
+  });
+
+  it(`hooks run with current param values`, async () => {
```
tdumitrescu, line 164:
```suggestion
  it(`runs hooks with current param values`, async function() {
```

lib/component-utils/paramsModule.js
```diff
@@ -0,0 +1,7 @@
+function updateParams(oldVnode, vnode) {
+  const elm = vnode.elm;
+  if (elm.setParams) {
+    elm.setParams(vnode.data.params || {});
```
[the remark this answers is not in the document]
hans-lizihan: i think what @themacguffinman suggested works as well

```
if (if (elm.setParams && vnode.data.params) {
```

lib/component.js
```diff
@@ -615,7 +636,7 @@ class Component extends WebComponent {
   }
 
   _render(state) {
-    if (this.shouldUpdate(state)) {
+    if (this.shouldComponentUpdate(null, state)) {
```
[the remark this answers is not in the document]
hans-lizihan: it won't. the base `Component` class's default `shouldComponentUpdate` will call `this.shouldUpdate` if param is null

https://github.com/mixpanel/panel/blob/hans-panel-params/lib/component.js#L251

lib/component.js
```diff
@@ -805,6 +826,89 @@ class Component extends WebComponent {
...
+      // set default value for the params
+      this.params[param] = paramSchemaObj.default;
+
+      this._paramSchemas[param] = paramSchemaObj;
+    }
+
+    return this._paramSchemas;
+  }
+
+  get params() {
```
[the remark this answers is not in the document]
hans-lizihan: ... more react way -> this.state vs this.setState 🙈

lib/component.js
hans-lizihan: 🤔 i think checking `must exist in param object` makes more sense, enforcing parent class to pass in the key

lib/component-utils/paramsModule.js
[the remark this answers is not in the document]
tdumitrescu: Yeah, I hadn't seen that at the time, looks like we hit upon a lot of similar points independently in our reviews. That being said, I do think it makes sense to learn from what works in the Snabbdom official modules, as those have been battle-tested for both efficiency and correctness.

lib/component.js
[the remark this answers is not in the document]
tdumitrescu: In that case why wouldn't we just leave this line as is. weird indirection we're introducing

lib/component-utils/paramsModule.js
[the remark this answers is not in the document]
hans-lizihan: right, i had the guards in initial PR and josh kinda want to move those checks in the `setParams` check

test/browser/index.js
tdumitrescu: But the tests in this repo already do use async/await, so what changed here that you need regenerator all of a sudden now?

lib/component.js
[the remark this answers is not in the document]
hans-lizihan: intend to enforce usage of `shouldComponentUpdate`, so child class should always use `shouldComponentUpdate`, if leave it with `shouldUpdate` then child class could never change this behavior with `shouldComponentUpdate`

test/browser/index.js
hans-lizihan: `babel-polyfill` had sth similar to `regenerator-runtime` built in.. but it has wrong behavior when it comes to `JSON.stringify(new Map())`

lib/component.js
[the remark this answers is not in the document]
tdumitrescu: I mean make it an internal method `_shouldUpdate` or something, but anyway this is bikeshedding over what ought to be temporary code, so don't worry about it.

test/browser/index.js
hans-lizihan: give https://github.com/mixpanel/panel/pull/153 a try :trollface:

lib/component.js
hans-lizihan: in that case I'm inclined to enforce `must exist` check behavior by default.

lib/component.js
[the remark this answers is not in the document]
hans-lizihan: i thought so as well, but @themacguffinman thinks it's better to keep it backward compatible 🙈

lib/component.js
hans-lizihan: oh interesting, i remember `for in` was like 5x faster than `for of` like 5 years ago. seems like v8 has optimized to a point there's no difference in terms of performance for those 2 usages now

lib/component.js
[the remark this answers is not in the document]
tdumitrescu: Yes, absolutely keep it backwards compatible. My suggestion was something slightly different, but really it's not worth spending cycles on, this version is fine, if a bit confusing.

lib/component.js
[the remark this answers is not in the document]
hans-lizihan: added guard to ensure `setParams` is only called if `vnode.data.params` is set

lib/component.js
hans-lizihan: second thought just enforce existence of param if explicitly specify `required: true`

test/browser/index.js
tdumitrescu: but you still use regenerator-runtime in that branch...

hans-lizihan: ah I recall, so still have some older browser tests. i could just bump browser version and specify browserlist correctly

tdumitrescu: yeah, that's what i mean about targeting a more modern environment

lib/component-utils/paramsModule.js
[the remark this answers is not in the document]
hans-lizihan: just added back those guards in params-module~

lib/component-utils/snabbdom-params-module.js
```diff
@@ -0,0 +1,13 @@
+function updateParams(oldVnode, vnode) {
+  const elm = vnode.elm;
+  // common guards used in snabbdom-attr-module
+  const oldParams = oldVnode.data.params;
```
hans-lizihan, line 4: added this guard just like attr module in snabbdom 

@themacguffinman, @tdumitrescu thinks that following snabbdom-attr-module's pattern to have some null check guards on the top may boost perf / more battle proved.

lib/component.js
```diff
@@ -276,6 +307,9 @@ class Component extends WebComponent {
 
     this.panelID = cuid();
 
+    // to enforce naviagate always point to this class
+    this.navigate = this.navigate.bind(this);
```
hans-lizihan, line 311: in case `navigate` need bo be passed in params by some component with routes

lib/component.js
```diff
@@ -805,6 +843,91 @@ class Component extends WebComponent {
...
+    this._paramSchemas = {};
+    const paramSchemas = this.getConfig(`params`);
+    const defaultParams = this.getConfig(`defaultParams`);
+
+    for (let [paramName, paramSchema] of Object.entries(paramSchemas)) {
+      // convert type shorthand to object
+      if (!paramSchema.type) {
+        paramSchema = {type: paramSchema};
+      }
+
```
hans-lizihan, line 862: removed check for `required` and `default` should not be presented same time since `require` now means the param should be declared in jsx, regardless what value it has

lib/component.js
```diff
@@ -805,6 +843,91 @@ class Component extends WebComponent {
...
+      // if param defined on schema and marked required, the key must be presented on the params
+      if (!params.hasOwnProperty(paramName) && paramSchema.required) {
+        throw new Error(
+          `param '${paramName}' on ${this.constructor.name} is defined as required param in schema but absent on component definition`,
+        );
+      }
+      const paramValue = params[paramName];
+      // set default value if empty value passed in
+      if (paramSchema.default && (paramValue === null || paramValue === undefined)) {
+        params[paramName] = paramSchema.default;
```
hans-lizihan, line 917: this could be a bit debatable, so if a value passed in as `null` or `undefined`, we will always give it `defaultValue`.

copied from react's behavior 
https://stackblitz.com/edit/react-ts-t7fty8?file=App.tsx

lib/component.js
```diff
@@ -805,6 +843,91 @@ class Component extends WebComponent {
...
+      }
+    }
+    const newParams = Object.freeze(Object.assign({}, params));
+
+    if (this.initialized && shouldComponentUpdate) {
+      this.runHook(`preUpdate`, updateOptions, null, newParams);
+    }
+    this._params = newParams;
```
hans-lizihan, line 925: now params get assigned regardless of shouldComponentUpdate returned value

lib/index.d.ts
```diff
@@ -295,3 +361,31 @@ export class Component<
...
+  StateT,
+  unknown,
+  unknown,
+  unknown,
+  ContextRegistryT,
+  ParamT
+> {}
+
+// define jsx IntrinsicElement inside namespace jsx to play well with react
+declare global {
```
hans-lizihan, line 378: this is a bit interesting, i found out by installing react and try to make snabbdom jsx and react jsx to play well together. after some trial and errors, it seems like the only way to do it is

1. put this global namespace in panel
2. since our jsxPragma is `jsx`, we need to make a namespace with `JSX.IntrinsicElements` inside it for typescript to correctly infer we are on snabbdom jsx.

hans-lizihan (commented):
Got some changes @themacguffinman @tdumitrescu 🙏

lib/component.js
themacguffinman: Interesting, I didn't expect that but I'm in favor of copying React's behavior.

lib/component-utils/snabbdom-params-module.js
themacguffinman: Fair enough, the attr module's guards seem more correct now that I took a second look at it. Sorry for leading you astray on this.

lib/component.js
themacguffinman: I suspect that this use case is rare enough that if a component needs to do this, it can use arrow functions by itself instead of binding in literally every component.

test/utils.js
```diff
@@ -1,3 +1,12 @@
+import serverNextAnimationFrame from './server/nextAnimationFrame';
+
 export function compactHtml(htmlStr) {
   return htmlStr.trim().replace(/>\s+</gm, `><`);
 }
+
+export function nextAnimationFrame() {
+  if (typeof requestAnimationFrame !== `undefined`) {
```
themacguffinman, line 8: Have you verified that you need this? `./server/nextAnimationFrame` already uses https://www.npmjs.com/package/raf which should already be a "polyfill for node and the browser"

lib/component.js
themacguffinman: I still think we should remove this binding.

themacguffinman (commented):
LGTM % comments, I'll leave final approval to @tdumitrescu

test/iso/param-component.js
```diff
@@ -0,0 +1,173 @@
+import sinon from 'sinon';
```
tdumitrescu, line 1: So this is the first thing in this "iso" dir, but I don't see any changes that would make CI run this in both browser and server env - what's the deal?

tdumitrescu (approved):
Looks pretty much good to go, just a few minor quibbles left 🎸 🎸 🎸

test/iso/param-component.js
hans-lizihan: the `browser` and `server` dir all import this file. 

https://github.com/mixpanel/panel/blob/hans-panel-params/test/browser/param-component.js
https://github.com/mixpanel/panel/blob/hans-panel-params/test/server/param-component.js

the original testcases are quite different in `browser` and `server` folder. but this testcase should be identical for both browser-test and server-test

tdumitrescu: that's an interesting approach, wouldn't it make more sense for the test commands to look in the `iso` dir rather than relying on files in both browser/ and server/ dirs to import from iso?

hans-lizihan: hmm, the `mocha` and `karma` commands target `browser` and `server` folder respectfully. 

mocha scans through the whole dir (test/server) whereas the karma relies on webpack build that has single entrypoint in `browser/index.js`

i think this is the minimum code to make it work on both test env

tdumitrescu: Whatever, it's not worth bikeshedding more, I meant adding `test/iso` to the `test-server` command in package.json and I guess adding to the webpack build for browser tests, feels like the right spots to configure this, but not a big deal either way

tdumitrescu:
🚀 published in v6.0.0

PR #165, opened 2022-10-03
Add support for styleOveride in params

hans-lizihan:


tdumitrescu (commented):
Do you think we should have any built-in typing for this? Or should each component that uses it have to declare an explicit entry for `styleOverride: String`?

hans-lizihan:
i think it's better to be explicitly declared, less magic

tdumitrescu (approved):
Fair enough

